PR #17, opened 2026-06-15
feat(itp): density-based convergence gate (tol_drho) + 60 μG diagnost…

Mi2ki510:
…ic trace

Add a gauge-invariant density convergence gate to the ITP loop and a standalone trace script that produces the GIF used to motivate it.

Why
---
Spinor F=6 ITP at low magnetic field sits in a near-degenerate manifold (polar pair m=±F). The wavefunction-based dpsi metric used previously counts U(1) global-phase drift and spin-z rotation drift as "non-convergence" even when the physical state is stationary. Eu-151 60 μG ITP runs reach a stable density profile by τ ≈ 100 but dpsi keeps floating at ~10⁻³, so save_every cadence affected the reported convergence verdict.

What
----
* `_run_itp_loop!` now computes per-(save_every) drho = max|Δρ|/max|ρ| alongside dE and dpsi, where ρ(r) = Σ_c |ψ_c(r)|² is the total density. ρ is invariant under both U(1) global phase and spin rotations about z, so the gauge-orbit drift that confused dpsi is removed.
* New `tol_drho` kwarg on `find_ground_state` (default 0.0 = ignore) gates convergence on `dE < tol AND drho < tol_drho` when set. Both metrics are still logged every save_every.
* GPU-compatible implementation: only broadcasting + reductions on the device array, no scalar indexing (the earlier in-loop drho rewrite was the one that allowed the gate to run on CUDABackend).
* Threaded through the YAML pipeline: `schema.jl` declares the field, `run_step_ground_state.jl` plumbs it.

Diagnostic evidence (not exercised by CI; documents the motivation) -------------------------------------------------------------------
* `scripts/flower_protocol_edh/itp_trace_60uG.jl` — 50 000-step ITP at B = 60 μG starting from m=+F, snapshots every 100 steps.
* `scripts/flower_protocol_edh/plot_itp_trace.py` — renders the trace as a polished GIF (xy/xz density + HSV phase + spin-arrow field coloured by |F_⊥| + populations bar + E/conv/spin time series with inset zoom of the last 30 %).
* `runs/eu151_flower_protocol_edh/main/submit_itp_trace_60uG.sh` — UGE submit for TSUBAME (gpu_h × 2 h).

The trace shows the canonical failure mode: from :m_plus_F init the state drifts through the spin manifold into an m=-F FM-like local minimum, drho ends ~10⁻⁴ at step 50 000 while dpsi is ~4·10⁻⁴ and the energy is still drifting upward — exactly the kind of "looks converged in ψ-metric but isn't physically stationary" case the new gate is intended to detect.

### DIFF
--- a/runs/eu151_flower_protocol_edh/main/submit_itp_trace_60uG.sh
+++ b/runs/eu151_flower_protocol_edh/main/submit_itp_trace_60uG.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# UGE submission: ITP trace at B=60 μG with on_step snapshot capture.
+# Outputs itp_trace_60uG.h5 → plot_itp_trace.py turns it into a GIF.
+#
+# qsub -g tga-kozuma-kouhi runs/eu151_flower_protocol_edh/main/submit_itp_trace_60uG.sh
+#$ -cwd
+#$ -N fpe_itp_trace
+#$ -l gpu_h=1
+#$ -l h_rt=2:00:00
+#$ -j n
+#$ -o /gs/bs/work/6/ue06186/bec-runs/flower_protocol_edh/logs/
+#$ -e /gs/bs/work/6/ue06186/bec-runs/flower_protocol_edh/logs/
+
+set -euo pipefail
+PROJECT_ROOT=$HOME/bec-simulation
+mkdir -p /gs/bs/work/6/ue06186/bec-runs/flower_protocol_edh/logs
+cd "$PROJECT_ROOT"
+source scripts/tsubame_setup.sh
+set -euo pipefail
+HOME_DEPOT=$HOME/.julia
+export JULIA_DEPOT_PATH="$JULIA_DEPOT_PATH:$HOME_DEPOT"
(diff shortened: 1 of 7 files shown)


anko9801:
Root-cause investigation of the 60 μG ITP behaviour that motivated this PR found that the `tol_drho` gate addresses a symptom rather than the cause: the ITP **fixed point itself** is displaced from the variational GP ground state (E=6.750 vs LBFGS 6.637, peak 0.0033 vs 0.0046), so it is not stationary under RTP (~49% breathing on handoff). A density convergence gate only changes *when* that wrong state is declared converged — it cannot make it the true ground state.

The actual bug is in the imaginary-time Euler Stage-3 (per-voxel `exp(-(m+F)·θ)` overflow shift with spatially-varying `θ ∝ |f(r)|`), fixed in #18. With #18 the ITP converges to E=6.644 (= LBFGS) and RTP breathing drops 49% → 7.4% (residual = benign finite-dt shadow).

Suggest revisiting this PR after #18 lands — the diagnostic trace/GIF tooling here is still useful, but the `tol_drho` convergence gate is likely unnecessary once the fixed point is correct.

PR #25, opened 2026-06-19
feat(ddi): spherically-truncated kernel (Ronen cutoff, Tier A)

anko9801:
## What

Opt-in **spherically-truncated DDI kernel** (Ronen–Bortolotti–Bohn cutoff; Vico–Greengard–Ferrando spectral form), threaded end-to-end through the YAML pipeline. Multiplies every k-space tensor component by the radial modifier

```
h(|k|R) = 1 + 3cos(x)/x² − 3sin(x)/x³,   x = |k|R
```

the Fourier transform of the dipolar kernel truncated in real space at radius `R`. This removes the lattice-anisotropy main term — the angular discontinuity of `k̂_α k̂_β` at the origin and the periodic-image tails of the un-padded convolution — with **spectral accuracy at fixed resolution**, instead of trying to converge it away by refining the grid.

## Why

The bare periodic kernel samples the `k=0` angular discontinuity on every grid, so residual C4-type anisotropy shrinks only slowly with resolution. The truncated kernel is `C^∞` (Paley–Wiener), so its grid samples converge spectrally — already near-isotropic at the working resolution.

## Convention preserved

- `h(0)=0` keeps the existing **`Q(k=0)=0` spherical-cavity** convention — **no contact / −δ term is introduced**. (Confirmed the current k=0 convention is "set to 0", not −δ, in both unpadded and padded paths.)
- `h(x)→1` for `|k|R ≫ 1` leaves bulk physics intact.
- **Off by default; bit-identical when disabled** (`× 1.0`).

## Surface

- `qtensor.jl`: `ddi_trunc_factor` (with small-`x` series guard) + `_build_q_tensor!(...; trunc_radius)`.
- `make_ddi_params` / `make_ddi_padded` accept `trunc_radius` (quasi-2D erfcx kernel exempt — already smooth).
- `make_workspace` gains `ddi_trunc_radius` (`NaN`=off, `≤0`=auto half-box, `>0`=explicit R), threaded through `find_ground_state` → ITP / LBFGS / adaptive / Jz, and the GS + dynamics pipeline steps.
- YAML: `ddi: {trunc_radius: <number> | "auto"}`; schema + reference docs updated.

## Tests / diagnostic

- `test/hamiltonian/test_ddi_truncated_kernel.jl` (registered in `FAST_TESTS`): pins `h(x)` limits/continuity/Float32, the `Q_trunc = h(|k|R)·Q_bare` relation, tracelessness, `k=0→0`, and Nyquist off-diagonal zeroing under truncation (A: code-correctness).
- `scripts/ddi_truncation_isotropy_probe.jl`: free-space/large-box convergence diagnostic (B: physics-fidelity).

## Note for reviewers

A naive *discrete* 45°-rotation symmetry test would be a false comfort — the bare grid kernel already respects every cubic point-group symmetry (incl. the xy-diagonal mirror) because `Q` is a function of the lattice-invariant `|k|²`. The genuine artifact (deviation from *continuous* rotation symmetry / periodic images) is only visible against a free-space reference, which is why that lives in the diagnostic script, not as a CI assertion.

⚠️ Not yet run under Julia (the dev container has no Julia toolchain) — relying on CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01VpncEMftyuEfqQkrTRCkrK)_


anko9801:
close: #24

PR #31, opened 2026-06-19
test: parallel runner (~2.1×) + orphan cleanup & structural tier enforcement

anko9801:
## Update — parallel test runner merged in (perf, ~2.1×)

This branch now also carries a **parallel test runner** on top of the orphan
cleanup below. Motivation: the per-push `fast` tier ran ~8 min serially (15 min
CI timeout). A new end-of-run per-file timing table showed the cost is
JIT-bound across ~130 unit files, not a few heavy outliers — so parallelism is
the only real lever.

**What landed**
- `SPINORBEC_TEST_WORKERS` (`1` = serial default, `N` / `auto` = parallel). CI
  `fast` → `auto`; nightly `full` → `2` (bounds peak memory).
- Parallel mode splits the slow-first file list into N chunks, each run in its
  **own independent julia process** (`test/run_chunk.jl`), aggregated by exit
  code. Independent processes — *not* Distributed workers — are deliberate: a
  shared worker pool reloads the package mid-run under a precompile-cache race
  and ends up with two copies of a type, so `x isa CheckResult` flaked to false
  (observed **2/3 runs red**). Separate processes load SpinorBEC exactly once
  and cannot — **10/10 green** after the switch.
- Serial and parallel share one run/fail/timing path (`test/_run_files.jl`), so
  they cannot disagree on what counts as a failure. Verified red-on-failure:
  an injected `@test false` / top-level `error()` turns **both** modes non-zero.
- `SPINORBEC_TEST_SKIP` for omitting env-specific files locally.

**Integration with the orphan-cleanup work below:** the tier lists +
`select_tests` moved to `test/_tiers.jl`, include()d by both `runtests.jl` and
`run_chunk.jl`, so `test_tier_membership.jl` (which reads `FAST_TESTS` /
`MANUAL_TESTS_ALLOWLIST`) works in the serial path *and* inside a parallel
chunk process — otherwise it errored `UndefVarError: FAST_TESTS` per chunk.

**Measured:** `fast` 204 s serial → ~98 s at 4 procs (~2.1×). Post-merge the
`fast` tier is green in both modes (133 files).

---

## What & why

Test-suite cleanup toward a leaner, **independent, meaningful, clean** structure. The guiding finding: after a full per-directory audit, the *running* suite is already well-curated — the physics-validation tests, the `oracles/` gates (CLAUDE.md commitment #3 "zero silent sign drift"), and the level0–12 validation ladder are load-bearing. The genuinely "half-baked" items were the **orphans** (files in no tier, running nowhere) and one **script masquerading as a test**, plus a double-execution coupling. This PR fixes those and makes the orphan state structurally impossible to recur.

**No working tests were deleted.** Cutting the oracle/physics/coverage spine would reduce exactly the physics-validation + coverage breadth the suite exists for.

### Orphan resolution (12 files were in no tier list)
- **→ CI_EXTRA** (load-bearing gates that ran nowhere):
  `hamiltonian/test_outer_operator_equivalence.jl` (ITP/RTP `V(dt/2)` chain equivalence),
  `oracles/test_gpu_cpu_fused_group_parity.jl` (fused-broadcast GPU/CPU axis; CUDA-gated, no-op on CPU),
  `oracles/test_multistart_winner_selection.jl` (stability-before-E ranking contract).
- **→ FAST_TESTS** (pure-introspection dispatch-coverage, fast + independent):
  `solvers/test_lbfgs_forward_coverage.jl`, `workflow/test_make_workspace_kwarg_coverage.jl`.
- **Kept manual** (GPU / heavy-YAML / TCP-server dependent): the 6 already in `MANUAL_TESTS.md`, now also enumerated in a new `MANUAL_TESTS_ALLOWLIST` so they read as *deliberately manual* rather than *orphaned*.
- **Relocated**: `test/validation/test_validation_matrix.jl` → `scripts/validation/run_validation_matrix.jl`. It's a CSV/Markdown-emitting runner, not a test. `@__DIR__/../..` still resolves to repo root, so the test paths it loads are unaffected; this also un-breaks the `docs/validation/*` links already pointing at the `scripts/` path.

### Independence
- `test/manuscript/test_f9_f11_polyhedral.jl` no longer `include()`s `test_lemma1_general_S.jl`. Both are FAST entries, so it ran twice and coupled the two files' pass/fail.

### Structure enforcement
- New FAST meta-test **`test/test_tier_membership.jl`** asserts: every `test_*.jl` under `test/` is in exactly one tier list or the MANUAL allowlist (no orphans), every listed path exists (no dangling), and nothing is double-listed across FAST/CI/FULL (no double execution). Makes CLAUDE.md commitment #7 self-enforcing.

## Verification note
This container has no Julia, so I verified statically (all referenced internal symbols still exist in `src/`/`scripts/`; orphan/dangling/double-listing sets are empty). CI runs the **fast** tier per-push (covers the meta-test + the 2 new FAST guards); the 3 new CI-tier tests are verified by nightly. If a newly-wired guard goes red, that's the guard doing its job (a real invariant drifted) — I'll follow up.

## Not done (deliberately)
Speculative LHY/DDI/TDHFB/workflow **merges** suggested by a broad first-pass scan were declined: deeper per-directory review showed each covers a distinct variant/path, so merging would *reduce* coverage breadth, and merged files can't be verified here without Julia. If you'd like to deliberately drop a whole subsystem layer (accepting the coverage tradeoff), say which and I'll do it as an explicit scope decision.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dx39xnKKAECHe8kYmfGGTk)_


anko9801:
### Update 2 — full-tier optimisation (LPT) + a pre-existing-failure finding

Extended the parallel runner to the **full (nightly) tier**:
- Round-robin chunking → **greedy LPT bin-packing** (`_balance` + a measured
  `_COST` table). Full-tier cost is skewed (a 161 s BO test, a 127 s
  triple-point, ~250 small units); LPT spreads the poles. Chunk file counts
  went `1/85/86/88` → `61/64/67/68`.
- nightly `full`: `SPINORBEC_TEST_WORKERS` 2 → **4** (measured peak ~2.7 GB
  per process → ~11 GB of 16 GB).
- Measured: full **846 s serial → ~450 s at 4 procs (~1.9×)**.

**Latent deps exposed + fixed.** Parallel chunks are fresh processes, so files
that used `norm`/`I`/`MersenneTwister`/… without a local import (they only
passed because the serial runner leaked the stdlib from a sibling) failed in
isolation. Fixed via a standard stdlib preamble in `_run_files.jl`
(LinearAlgebra/Random/Statistics/Printf — the environment contract both modes
share) plus explicit imports in the offending files. Also fixed
`oracles/test_path_coverage.jl`, which scraped `runtests.jl` for the tier list
that now lives in `_tiers.jl`.

**Verified outcome-neutral:** the parallel failure set == the serial set.

> ⚠️ **Pre-existing full-tier failures (NOT from this PR).** Running `full`
> locally surfaced ~8 failing files that fail identically in serial and are
> unrelated to the runner: `oracles/test_propagator_references.jl` (a real
> `_update_batched_kinetic_phase!` **MethodError** → cascading Strang-slope
> fails), `test_level4_general_F_phase_emergence.jl`, the F32
> `gpu/test_mixed_precision*` (CPU path), `rotating_basis/test_scalar_egpe_*`,
> `test_rotating_basis_phase_ii.jl`, `analysis/test_spatial_zeeman.jl`. These
> were added in older commits, run nightly-only, and have drifted red (the
> MethodError is environment-independent; some others may be FP/precision
> tolerances). Worth a separate look — this PR neither causes nor fixes them.

PR #62, opened 2026-07-05
fix(rotating_basis): GPU-enable + rotating-field EdH/Barnett study

anko9801:
## What

Two things:

1. **Framework fixes** that let `kind: rotating_basis` run on GPU again (3 files, `src/`).
2. **A reproducible rotating-field EdH + Barnett study** (Eu-151 F=6 dipolar BEC): scripts, YAML configs, README, key figures under `runs/eu_barnett_rotfield_clean/`.

## Framework fixes (the reviewable code)

| file | fix |
|---|---|
| `hamiltonian/coefficients.jl` | Restore `compute_gamma_lhy` (dropped in the RotatingBasisWS-retirement refactor). The rotating_basis GS auto-path calls it for `ε_dd > 0.5` — i.e. every Eu-151 run — giving `UndefVarError`. Reinstated with the same Lima-Pelster formula the non-rotating resolver uses. |
| `analysis/currents.jl` | `orbital_angular_momentum` scalar-indexed a GPU `psi` (crash under the rotating_basis per-save `⟨L_z⟩` poll). Bring `psi` host-side once + cache CPU FFT plans by shape (stop rebuilding them every poll). |
| `workflow/experiments/inspect_checks.jl` | Downgrade the `rotating_omega_spinor_gpu` compat rule `:block → :warn`. The underlying Coriolis-shear GPU bug was fixed 2026-06-02; the hard block is stale. |

After these, `rotating_basis` magnetostir completes on GPU (32³, ~2.5 min; dynamics stepping ~75–86% util). The 48³ run is heavy (secular DDI + Yoshida-4 ≈ 100 ms/step) but genuinely GPU-bound, not a fallback.

## Physics study (`runs/eu_barnett_rotfield_clean/`)

- **Einstein-de Haas**: field quench → spin→orbital transfer, `J_z = F_z + L_z ≈ 6` conserved, real vortices (density holes + 2π winding) in transferred m-components (`edh_baseline.png`).
- **Direction control** from a transverse (`J_z=0`) start: `⟨L_z⟩(+Ω) = −⟨L_z⟩(−Ω)` and `⟨F_z⟩(+Ω) = −⟨F_z⟩(−Ω)` exactly; `Ω=0` control does nothing.
- **DDI on/off control** (`master_figure.png`, `ddi_control.png`): vortices **require** the DDI (`⟨L_z⟩=0` without it, `|F|` stays 6). The raw `⟨F_z⟩` oscillation alone is mostly forced Larmor precession — the genuine many-body signal is the vortex generation + depolarisation.
- **Optimization** (`optimization_scaling.png`): response vs rotation rate Ω and rotating-field amplitude.

Analysis tooling recomputes `⟨L_z⟩/⟨J_z⟩` + a winding-based vortex census + spin vector from ψ snapshots (the lab-frame spinor save path stores `F_z` but not `L_z`).

Large outputs (mp4 videos, per-frame CSV dumps, jld2 results) are intentionally left out of git; regenerate via the `run_*.jl` drivers + `make_video_*`.

## Test notes

- rotating_basis GS ITP + magnetostir dynamics verified on GPU (WSL2, 5070 Ti).
- No behaviour change for existing CPU rotating_basis runs; `orbital_angular_momentum` CPU path unchanged except plan caching.


anko9801:
Closing as superseded — but not before checking what would be lost, since this branch is 219 files and three weeks stale.

**Superseded by #141**, which is a clean redo of `runs/eu_barnett_rotfield_clean/` on corrected code. That redo also fixes two things this branch could not have known about: the ±Ω arms here were not mirror images (`SinusoidalWaveform` is **sin**, so `phase_x = ±π/2` gave B(0) = ±x̂ and the +Ω arm started 180° anti-aligned with the ground-state spin), and `run_step_rotating/ground_state.jl` was calling a `compute_gamma_lhy` that no longer exists.

**What was salvaged first.** Diffing the five `src`/`test` files here against both `main` and #141, four were purely behind main. One was not: `src/analysis/currents.jl` carried a real fix that exists nowhere else — `orbital_angular_momentum` **threw** on a device array (`Scalar indexing is disallowed`), which matters because ⟨L_z⟩ is the observable the EdH/Barnett J_z ledger is written in and those runs are on GPU. That is now in `main` via #147, with a CUDA-gated `device == host` equality test.

One correction while salvaging it: this branch described FFT-plan rebuilding as *"the dominant cost"* of that observable. Measured, it is 1.05–1.10× (`FFTW.ESTIMATE` plans are cheap) — kept, but not as the headline.

Nothing else here is unique to this branch.

PR #99, opened 2026-07-27
feat(analysis): superfluid fraction from the phase-twist free energy

anko9801:
Adds `superfluid_fraction(psi_or_n, grid; direction, method)` — the translational $f_s$ along one axis, from the energy cost of twisting the phase across the periodic box ($\hbar = m = 1$).

## Two branches, gated against each other

- **`:leggett`** — plane-average closed form $f_s = L^2/[(\int \bar n)(\int dx/\bar n)]$, the harmonic-over-arithmetic mean of the density. Cheap, and the quantity the supersolid literature reports.
- **`:relaxed`** — the full variational minimum over a periodic phase: a finite-volume solve of $\nabla\cdot(n(\hat e_d + \nabla u)) = 0$ by CG on the singular-but-compatible system.

They agree to discretisation error whenever the density is modulated along the flow axis only — that is what gates one against the other. Elsewhere `:relaxed ≤ :leggett`, because letting the phase vary transversally lets the flow concentrate where the density is high, which lowers the cost.

## Why it isn't a tautology

The naive "boost the condensate and read off the kinetic energy" estimator returns 1 for *any* state, since $|\psi|^2$ is unchanged by the boost. Holding the density rigid while relaxing the phase is what makes the number informative — and is also why both branches are upper bounds on the true $f_s$.

## Validation (type A only, deliberately)

- Analytic oracle: a cosine density of contrast $A$ gives $f_s = \sqrt{1-A^2}$ exactly.
- The relaxed branch converges to that at second order in $dx$ (pinned by a refinement test).
- Bounds $0 \le f_s^{\text{relaxed}} \le f_s^{\text{leggett}} \le 1$ over arbitrary densities.
- Transverse-channel case where the two branches must *differ*, so the agreement test above cannot pass vacuously.

No type-B or type-C claim is made here. The rigid-density approximation is untested against a twisted-BC re-minimisation, and nothing has been compared to a published $f_s$ — both are follow-ups.

## Known limits

- CPU only: the `CartesianIndices` loops scalar-index, so a GPU ψ needs `_to_host` at the entry. Not yet wired through `_run_analyzer` (which already hosts ψ), so this lands with the analyzer follow-up.
- The CG is unpreconditioned and unbenchmarked at production grid sizes.
- A trapped cloud with vacuum at the box edge legitimately gives $f_s \approx 0$ (no phase rigidity across the box); that case warns rather than silently returning zero.

## Tests

`analysis/test_superfluid_fraction.jl` 42/42, registered in the fast tier. Full fast tier 142/142.

## References

- A. J. Leggett, *Can a solid be superfluid?*, Phys. Rev. Lett. **25**, 1543 (1970), doi:10.1103/PhysRevLett.25.1543
- G. Biagioni et al., *Measurement of the superfluid fraction of a supersolid by Josephson effect*, Nature (2024), doi:10.1038/s41586-024-07361-9

### DIFF
--- a/src/analysis.jl
+++ b/src/analysis.jl
@@ -8,6 +8,8 @@
 #       spin observables, ensemble stat helpers
 #   energy/currents/vorticity     — energy decomposition, probability
 #       current, superfluid velocity/vorticity
+#   superfluid_fraction            — phase-twist f_s (Leggett plane-average
+#       bound + full variational relaxation)
 #   vortex_extraction              — per-m vortex line tracing (3D)
 #   diagnostics                    — Zeeman/healing-length/TF radius helpers
 #   sinatra_diagnostics/grid_resolution — TWA-validity (Sinatra) per-knob
@@ -27,6 +29,7 @@ include("analysis/ensemble.jl")
 include("analysis/energy.jl")
 include("analysis/currents.jl")
 include("analysis/vorticity.jl")
+include("analysis/superfluid_fraction.jl")  # phase-twist f_s (Leggett bound + relaxed)
 include("analysis/vortex_extraction.jl")
 include("analysis/diagnostics.jl")
 include("analysis/sinatra_diagnostics.jl")  # TWA-validity (Sinatra) per-knob helpers
--- a/src/analysis/superfluid_fraction.jl
+++ b/src/analysis/superfluid_fraction.jl
@@ -1,0 +1,246 @@
+export superfluid_fraction, plane_averaged_density
+
+# Translational superfluid fraction from the phase-twist free energy.
+#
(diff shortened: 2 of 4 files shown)


anko9801:
Correction from the stacked follow-up #105: this PR's body and docstring say both branches are "upper bounds on the true superfluid fraction". That caveat is in the wrong place.

Within Gross–Pitaevskii mean field the rigid-density value is **exact**: the density response to the twist is $O(q^2)$, and since the untwisted state is stationary it only reaches the energy at $O(q^4)$, leaving $f_s$ (an $O(q^2)$ quantity) unaffected. Verified in #105 against a direct twisted-BC GP minimisation across $f_s$ from 1.5e-4 to 0.53.

The bound is only a bound beyond mean field, where a one-body phase cannot represent correlated backflow. #105 corrects the docstring.

PR #104, opened 2026-07-27
feat(analysis): reach superfluid_fraction from YAML, Experiment and GPU

anko9801:
Stacked on #99.

The metric landed in #99 with **no route into a production run** — no `analyze:` entry, no `Experiment` observable, and it threw on a device array. This adds the three faces, plus one discretisation fix the new tests forced out.

## Reachability

- **`superfluid_fraction` analyzer** (`spectroscopy.jl` + `_run_analyzer` branch). Per-axis; `method` ∈ `{leggett, relaxed, both}`; the unselected branch returns `NaN` rather than a silent `0`.
  ```yaml
  analyze:
    - superfluid_fraction: {}                          # all axes, both methods
    - superfluid_fraction: {directions: [1], method: leggett}
  ```
- **`superfluid_fraction(exp)` / `(exp, t)`** observables. Grid comes from the run's own `RunResult`, so the value is reproducible from the jld2 alone.
- **Device arrays.** Every loop scalar-indexes, so the entry point hosts first. The density face also widens `Float64` → `Real`, which the mixed-precision (`dtype: f32`) path needs; the solve still runs in `Float64`.

## Discretisation fix: face densities stay arithmetic

Harmonic face averaging was tried. It is the textbook choice for a discontinuous coefficient, and it makes $\sum 1/n_{\text{face}} = \sum 1/n$ hold exactly — but it breaks $\sum n_{\text{face}} = \sum n$, and *that* is the identity the $f_s$ ratio needs.

The disconnected-cloud test caught it immediately: a 21-cell blob in vacuum has 20 harmonic faces against 21 cells of mass, and reports $f_s = 1/21$ instead of the correct 0. Arithmetic faces supply the two half-weight edge faces that close the sum. Both the reasoning and the counterexample now live in the source comment so the "obvious" harmonic switch does not get made again.

## Corrected claim: the branch ordering is a continuum statement

That same experiment showed the ordering claim in #99 was too strong. `:leggett` is a spectrally accurate trapezoid sum; `:relaxed` is second-order with a **one-sided positive** truncation error. So at coarse `dx` the relaxed branch legitimately sits slightly *above* the bound — 1e-4 at 128 points / contrast 0.5, 8e-3 at 64 points / contrast 0.99.

A test now pins both the sign and the order of that gap, and the docstring carries the numbers, so a user can tell discretisation from transverse rerouting when the two branches disagree. Without this, someone would have read a coarse-grid excess as physics.

## Measured before optimising

`:relaxed` at 128³ is **1.8 s**, and the cost is flat in density contrast (0.21 s → 0.25 s going from contrast 0.2 to 0.99 at 64³). The FFT preconditioner this was expected to need is **not needed** — the planned optimisation is cancelled rather than written.

| grid | `:relaxed` |
|---|---|
| 32³ | 0.006 s |
| 64³ | 0.25 s |
| 96³ | 0.79 s |
| 128³ | 1.80 s |

## Tests

`analysis/test_superfluid_fraction.jl` 55/55 · `gpu/test_superfluid_fraction_gpu.jl` 7/7 on a live device (F64 and F32 paths) · tier membership 33/33 · fast tier 142/142.

### DIFF
--- a/docs/reference/yaml_schema_reference.md
+++ b/docs/reference/yaml_schema_reference.md
@@ -500,6 +500,7 @@ A list of analyzers, each `{<name>: <params>}`. Names include:
 - `population_history`
 - `trap_population` (radius, center) — inside-vs-outside-trap norm split; tracks atoms spilling toward the absorbing boundary
 - `cloud_shape` — center of mass, per-axis RMS widths, principal-axis widths, aspect ratio, in-plane tilt; quantifies cloud deformation across a snapshot series
+- `superfluid_fraction` (directions, method) — per-axis $f_s$ from the phase-twist free energy: `leggett` (plane-average bound), `relaxed` (full variational minimum), or `both` (default). Rigid-density, so both are upper bounds; a cloud that does not span the periodic box legitimately reports ≈ 0
 
 Full list in `src/workflow/experiments/analyzers/`.
 
--- a/src/analysis/superfluid_fraction.jl
+++ b/src/analysis/superfluid_fraction.jl
@@ -13,10 +13,11 @@ export superfluid_fraction, plane_averaged_density
 #
 #     f_s = min_u ∫ n |ê_d + ∇u|² dV / ∫ n dV
 #
-# The density is held rigid, which makes f_s an upper bound on the true
-# superfluid fraction. Restricting u to depend on x_d alone gives Leggett's
-# closed form — a further upper bound, and the quantity usually reported for
-# supersolids.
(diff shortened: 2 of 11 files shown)


anko9801:
Superseded by #121 — same content plus a merge of current `main`. This PR was auto-closed by GitHub when its base branch (`feat/superfluid-fraction`) was deleted during the merge of #99; it was not merged, and its content had not reached `main`. #121 targets `main` directly and carries the #105 / #106 commits too, since those merged into this branch rather than into `main`.

PR #108, opened 2026-07-27
fix(lhy): auto-derived scalar c_lhy was short by π·(a_s/a_ho)·√N

anko9801:
Found while checking whether the scalar eGPE path could produce a dipolar supersolid — the LHY strength sets the droplet scale, so it had to be verified before use.

## The bug

`_resolve_lhy_block!` filled `c_lhy` with $(128/(3\sqrt{\pi}))\,\tilde a^{3/2} N$. For this repo's per-particle convention ($\int|\psi|^2 dV = 1$, $E_{LHY}/N = \frac{2}{5} c_{lhy}\int n^{5/2}dV$, $c_0 = 4\pi\tilde a N$) the correct value is

$$c_{lhy} = \frac{128\sqrt{\pi}}{3}\,\tilde a^{5/2} N^{3/2}\, Q_5(\epsilon_{dd}), \qquad \tilde a = a_s/a_{ho}$$

Both the $a_s$ exponent and the $N$ exponent were low. The shipped value is too small by $\pi\tilde a\sqrt{N}$ — **11.7× for ¹⁶⁴Dy** at $N = 6\times10^4$, $\omega = 2\pi\cdot600$ Hz; **6.8× for ¹⁵¹Eu** at $N = 5\times10^4$, 110 Hz.

## No convention freedom

$\gamma_{QF} = \frac{32}{3} g\sqrt{a_s^3/\pi}$ with $g = 4\pi\hbar^2 a_s/m$ fixes the ratio of the two chemical-potential terms in SI:

$$\frac{\mu_{LHY}}{\mu_{contact}} = \frac{32}{3}\sqrt{n_{SI} a_s^3/\pi}$$

and the dimensionless form must reproduce it. Measured at three densities across Dy164 / Er168 / Eu151: old coefficient gives 0.0334 where SI says 0.3902; new one matches to 1e-12.

## Why it survived

The error is dimensionless-to-dimensionless, so no units check could see it, and **no test pinned the derived number against anything physical** — every existing LHY test passes `c_lhy` in explicitly. The new oracle drives the real code path (`_resolve_lhy_block!`, not a re-typed formula) and gates the SI ratio *plus both scaling exponents separately*, since getting either one wrong is exactly the failure mode that shipped.

## Blast radius

Fires only where a config writes `lhy: {kind: scalar}` (or `quasi_2d`) with no explicit `c_lhy` and `c_dd > 0`. **About 60 YAMLs under `runs/` are in that set**, including `eu151_edh`, `eu_k3_arrest`, `eu_robust_factorial`, `saito_li_torus` and the TWA ε_dd / N scans — their LHY term was ~7–12× too weak.

Configs that set `c_lhy` explicitly are untouched; a test pins that the explicit value is never overwritten.

⚠️ Results from the affected runs should be re-checked before being quoted. I have not re-run them.

## Tests

`test_scalar_lhy_si_roundtrip` 16/16 (new, fast tier) · `test_lhy` 37/37 · tier membership 33/33.

### DIFF
--- a/src/workflow/experiments/schema/parsing_blocks.jl
+++ b/src/workflow/experiments/schema/parsing_blocks.jl
@@ -323,8 +323,19 @@ function _resolve_lhy_block!(p::Dict, inter::Dict, atom, c_dd_val::Float64,
     # Auto-derive c_lhy for scalar / quasi_2d when user omitted it.
     if (kind == "scalar" || kind == "quasi_2d") && !haskey(lhy_block, "c_lhy")
         if !isnan(c_dd_val) && c_dd_val > 0
-            a_s_dl = atom.a_s / a_ho
-            c_lhy_scalar = (128.0 / (3.0 * sqrt(π))) * sqrt(abs(a_s_dl)^3) * N_atoms
+            # Dimensionless scalar LHY coefficient for the repo's per-particle
+            # convention (∫|ψ|²dV = 1, E_LHY/N = (2/5)·c_lhy·∫n^{5/2}dV,
+            # c₀ = 4π(a_s/a_ho)N). Fixed by SI alone — no convention freedom:
+            #
+            #   γ_QF = (32/3)·g·√(a_s³/π),  g = 4πℏ²a_s/m
+            #        = (128√π/3)(ℏ²/m)·a_s^{5/2}
+            #   ⇒ c_lhy = (128√π/3)·(a_s/a_ho)^{5/2}·N^{3/2}·Q₅(ε_dd)
+            #
+            # equivalently c_lhy/c₀ = (32/3)√((a_s/a_ho)³N/π), which reproduces
+            # the textbook ratio μ_LHY/μ_contact = (32/3)√(n_SI·a_s³/π).
+            # test/oracles/test_scalar_lhy_si_roundtrip.jl gates exactly that.
+            a_s_dl = abs(atom.a_s / a_ho)
+            c_lhy_scalar = (128.0 * sqrt(π) / 3.0) * a_s_dl^2.5 * float(N_atoms)^1.5
             lhy_block["c_lhy"] = c_lhy_scalar * lima_pelster_Q5(eps_dd)
         end
     end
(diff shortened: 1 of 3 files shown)


anko9801:
### Triage: how much does this actually move the physics?

Two corrections to the PR body above.

**1. "Too weak" is only true above a crossover.** The error factor is $\pi\tilde a\sqrt N$, which passes through 1 at $\sqrt N = 1/(\pi\tilde a)$ — about $N \approx 1100$ for ¹⁵¹Eu at $\omega_{ref} = 691$ rad/s. Below that the old coefficient was too **large**. So the bug is not uniformly one-signed; it is $N$-dependent, which is exactly what an $N$ vs $N^{3/2}$ exponent error looks like.

**2. In absolute terms the LHY term is small either way for the affected Eu runs.** Estimating the peak density from the isotropic Thomas-Fermi profile ($\mu = \frac12(15c_0/4\pi)^{2/5}$, $n_{peak} = \mu/c_0$):

| family | N | LHY/contact (fixed) | (old) |
|---|---|---|---|
| `eu151_edh` / `barnett` / `twa` | 1e4 | 3.6 % | 1.56 % |
| `twa_N_scan` | 1e3 | 2.3 % | 3.10 % |
| `twa_N_scan` | 1e5 | 5.8 % | 0.78 % |
| `saito_li_torus` | 1.5e4 | 4.0 % | 1.38 % |
| `eu_k3_arrest` / `eu_robust_factorial` / `l4_k3_ladder` | 3e4 | 4.3 % | 1.11 % |
| `eu151_klaus_*` | 6e4 | 3.3 % | 0.84 % |

So no affected Eu run had its mean field dominated by the LHY term before or after. The correction is a few percent of the contact energy, not a regime change.

**Where it still matters.** The runs designed to *measure* LHY sensitivity — `eu_k3_lhy`, `eu_k3_lhy_control`, and the LHY0/LHY1 arms of `eu_robust_factorial` — were comparing against a term ~4× weaker than intended. A "LHY makes no difference" verdict from those is not automatically wrong, but it was tested at the wrong strength and should be re-run before being quoted. `twa_N_scan` at N = 1e5 has the largest relative change (0.78 % → 5.8 %).

The ¹⁶⁴Dy numbers in the PR body stand and are the reason this surfaced: dipolar supersolids sit where LHY is order-unity, so the coefficient has to be right before that work can start.

PR #119, opened 2026-07-27
feat(lhy): warn when a textured state meets a single-spinor LHY table

anko9801:
Every spinor LHY table is built for **one** spinor and then applied at every voxel — `:full_bdg` takes the peak-density spinor via `_extract_spinor`, and the closed forms assume their own fixed ansatz without looking at the state at all. Exact when the state is uniform, an approximation otherwise, and nothing said so.

I measured it before deciding what to do about it.

## Only the magnitude of ⟨F⟩ matters

`ε_LHY` is an SO(3) scalar for contact, so **rotating** the spinor leaves it invariant to machine precision across the whole canting family; the DDI breaks that by **0.25%** at ε_dd ≈ 0.05. Varying `|⟨F⟩|/F` from 1 to 0 costs **~20%**.

So every direction texture in `state_zoo` — flower, radial/chiral spin vortex, skyrmion, all `|⟨F⟩|/F = 1` everywhere — is **free**. That is why this has never bitten.

## Converged states are a different matter

On the pinned weak-field Eu B-scan (`figs/eu_bscan_pin_tight`), `|⟨F⟩|/F` develops real spatial structure as B falls, and the peak value diverges from the `n^(5/2)`-weighted mean the energy actually carries (**0.640 vs 0.799** at frame_046). Shipped single-spinor table vs a per-voxel LDA over the full weight:

| `\|⟨F⟩\|/F` spread | ε_LHY error |
|---|---|
| 0.00 | +0.00% |
| 0.375 | −1.48% |
| 0.892 | **−4.48%** |
| 0.901 | **+4.87%** |

~5%, and **the sign flips along the scan** — so it does not cancel in a B-comparison, which is the worst shape for a scan hunting a transition. That is what makes it worth surfacing rather than leaving as a docstring caveat.

## What this does *not* do

It does not build spatially-varying LHY. 5% does not justify that redesign, and **no config in the repo currently combines a spinor LHY kind with a textured state** — all 20 use `initial_state: m_minus_F`, uniform. This is a **latent trap, not a live error**.

But the active weak-field Eu arcs are exactly the textures that would spring it, and the original framing of the LHY work called this the common bottleneck. So `make_workspace` warns above a `|⟨F⟩|/F` spread of 0.3 (the threshold is where the error leaves the noise, from the table above), naming the measured cost, and CLAUDE.md records the limitation.

## Method note, because the first answer was wrong

The first pass sampled the **top 140 voxels by `n^(5/2)` weight** and reported 0.5%. That covers 7–9% of the weight and — worse — selects precisely the uniform high-density core while missing the varying halo. **Biased toward "no error" by construction.**

The coverage figure was printed alongside the result, which is how it was caught. Redone by binning on `|⟨F⟩|/F` so 100% of the weight is covered; the within-bin spread (~2%) is reported as the method's own uncertainty rather than hidden.

## Tests

`test/workflow/test_lhy_texture_warning.jl` (27): direction textures stay silent, a magnitude texture is detected and reported (both for `:full_bdg` and for a closed form, with wording that names the ansatz), degenerate inputs abstain rather than mis-measure, and the warning is reachable through a real `make_workspace` call rather than only from the helper.

Fast tier: 145 files, all passed.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -250,6 +250,7 @@ NOT bugs — don't "fix".
 - **`spin_rotating_frame_omega ≠ 0` requires `secular_ddi=true`** (enforced via `ArgumentError`). Full DDI's off-diagonal components Larmor-average to zero only in secular limit.
 - **`even_c_extra(F; c2, c4, c6, …)` is canonical** — hand-written `[c2, c4, c6]` silently misindexes for F ≥ 3.
 - CUDA-Graph in `ext/SpinorBECCUDAExt/gpu_graph.jl` disabled (replay drift, 4× slower); `split_step_captured!` itself was deleted 2026-05-22.
+- **Spinor LHY tables are built for ONE spinor and applied at every voxel** — `:full_bdg` takes the peak-density spinor, the closed forms assume their own fixed ansatz. Exact for a uniform state; measured ~5% in ε_LHY on converged weak-field Eu textures, with a sign that FLIPS along a B-scan (so it does not cancel in a comparison). Only the magnitude of ⟨F⟩ matters — a pure direction texture (flower / spin vortex / skyrmion, all |⟨F⟩|/F = 1) is free, since ε_LHY is an SO(3) scalar for contact and moves 0.25% under the DDI. `make_workspace` warns above a |⟨F⟩|/F spread of 0.3 (`test/workflow/test_lhy_texture_warning.jl`). Spatially-varying LHY is not implemented.
 - **Tensor c2/c4 not in `energy_gradient!`** — LBFGS warns, falls back to ITP. `[KNOWN-LIMIT]`.
 - **TDHFB has no YAML pipeline integration.** `dynamics.tdhfb` does NOT exist. Engine parallel-track to GP; do not wire in without explicit ask.
 
--- a/src/workflow/initialization/make_workspace.jl
+++ b/src/workflow/initialization/make_workspace.jl
@@ -560,12 +560,18 @@ rotating the spinor leaves ε_LHY invariant to machine precision for contact
 direction texture (flower, spin vortex, skyrmion; all `|⟨F⟩|/F` = 1 everywhere)
 costs nothing. Varying the MAGNITUDE costs ~20% between `|⟨F⟩|/F` = 1 and 0.
 """
-function _lhy_texture_spread(psi_init, F::Int)
-    psi_init === nothing && return (0.0, 1.0, 1.0)
-    N = ndims(psi_init) - 1
-    D = size(psi_init, N + 1)
+_lhy_texture_spread(::Nothing, ::Int) = (0.0, 1.0, 1.0)
+
+# `Val(N)` comes from the ARRAY'S type parameter, never from `ndims(x)` —
+# CLAUDE.md "Type stability boundaries". Taking `N = ndims(psi_init) - 1` as a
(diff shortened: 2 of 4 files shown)


anko9801:
Pushed `911a07ed` — **the guard cost more than everything it guards**, found by measuring it before claiming it was ready to merge.

```
96³:  guard 0.314 s / 695 MB   vs   the closed-form table build 0.00004 s
```

Two causes, both mine:

- **It restated existing physics.** `spin_density_vector` already computes ⟨F⟩ in the O(D) ladder form (F± tridiagonal, Fz diagonal), threaded and gated. I was doing 13×13 matrix-vector products per voxel, allocating a fresh vector each time.
- **It swept every voxel to answer a yes/no threshold question** — inside `make_workspace`, which CLAUDE.md calls the hot path for every pipeline step and which runs per scan point.

Now built on the existing helper and subsampled on a stride targeting ~8000 points. Against the **right denominator** — the `make_workspace` call it runs inside, not the sub-step — that is **26% → 5% at 64³**. Below ~20³ the stride is 1 and nothing is skipped.

Verdict unaffected: the converged weak-field Eu state this was calibrated on reads a spread of **0.77 sampled against 0.92 full-grid**, both far above the 0.3 threshold. The reported peak/mean move a few percent (0.699/0.807 vs 0.640/0.799), so the warning now says *sampled, indicative not exact* and rounds to two digits instead of implying precision it does not have.

Two corrections to my own numbers along the way, both the same shape as the sampling bias already described in the PR body:

- **the 40000× was the wrong denominator.** The guard runs once per workspace, so the workspace is what it should be compared against. It reads alarming; 26% is the figure that actually mattered.
- **I wrote "spread 0.921 either way" into a comment before measuring it.** It is 0.770. The verdict is unchanged but the number was not, and the warning quotes these numbers to the user.

Fast tier: 145 files, all passed.

anko9801:
Pushed `9180cab4`. **This was not safe to merge before it** — found by checking type stability when asked whether it was ready.

`_lhy_texture_spread` inferred as `Tuple{Any,Any,Any}`, on the `make_workspace` path CLAUDE.md singles out:

> `Workspace` has 23+ type parameters; `Dict{Symbol,Any}` or closures escaping into Workspace paths cause inference to explode (multi-minute JIT hang, no stack trace).

**Tests were green and CI was green. Neither detects this.** The symptom is not wrongness — it is someone later hitting an unexplained multi-minute hang.

The cause was the documented rule violated directly: *"`Val(N)` from a type parameter, not `Val(ndim::Int)`"*. I took `N = ndims(psi_init) - 1` as a runtime Int, which left `ntuple(..., N)` uninferrable and poisoned the return type. Now dispatching on the array's own `M` parameter with `Val(N)`, and narrowing `spin_density_vector`'s result at the call boundary (`spin_matrices(F)` is `SpinMatrices{D}` with `D` only known at runtime).

`@inferred` is asserted for 1D/2D/3D and the `nothing` input, so it cannot come back quietly — inference is part of this function's contract given where it sits, not a nicety.

---

That is the **third** correction to this guard, and all three surfaced only because I was asked whether it was ready rather than assumed it:

1. the LDA sampling covered 7–9% of the weight and selected the uniform core — biased toward "no error" by construction
2. the guard cost more than everything it guards (restated `spin_density_vector`; full-grid sweep on a hot path)
3. this one

Each was visible to a check I had not run. "CI is green" caught none of the three.

Fast tier: 145 files, all passed.

PR #125, opened 2026-07-28
fix(lhy): tabulated LHY was silently dropped on the broadcast propagator path

anko9801:
**Every GPU run using a tabulated LHY was running with no LHY at all**, and nothing said so. 12 configs under `runs/` are affected — including all six `eu_lhy_longtime` points and the `LHY_{polar_contact,icosahedral}_*ms` series, i.e. runs named after the term they were not applying.

Found while checking whether #124 was production-ready. It is a pre-existing bug, older and more serious than that PR.

## The bug

The generic `_diagonal_step_svec!` summarised the LHY as a single coefficient:

```julia
c_lhy_val_t = c_lhy isa Float64 ? c_lhy : (c_lhy isa ScalarLHY ? c_lhy.c_lhy : 0.0)
```

That is only the shape of `ScalarLHY` (`c·n^{3/2}`). Every `TabulatedLHY` fell to `0.0` — while `_has_lhy` still read `true`, so the branch *ran*, with the LHY removed.

```
PolarContactLHY:  max|ψ_fused − ψ_generic| = 5.7 after ONE step
                  V_LHY(n=1) = 50.5, simply absent
```

**Reproducible on CPU alone** (a `view` forces the same generic method), so this is not really a GPU bug — the GPU just always takes that path, because the fused kernel's `c_lhy` bound admits only `Nothing`/`NoLHY`/`Float64`/`ScalarLHY`.

Affected: `polar_contact`, `fm_contact`, `icosahedral`, `polar_dipolar`, `fm_dipolar`, `polar_two_channel`, `full_bdg`. CPU runs through the fused `::Array` kernel are correct and unaffected.

## Fix

The broadcast path now materialises **V_LHY as a field** via `_lhy_potential_field` instead of reducing it to a coefficient. Both paths route through `_lhy_V`, so they cannot disagree about what the LHY *is*.

The CUDA extension gets a device method (`gpu_lhy_field.jl`): the table uploads once per `objectid` and is looked up with O(1) index arithmetic. That is valid because every tabulator builds `range(0, n_max; length=n_points)` — **checked at the call, not assumed**, since a non-uniform grid would be silently mis-indexed, which is the same class of failure as the bug being fixed.

Verified **GPU == CPU to ≤ 7.7e-15** across `polar_contact` / `fm_contact` / `icosahedral` / `polar_two_channel` / `ScalarLHY` / `NoLHY`, real and imaginary time.

## Why the existing oracle missed it

There is a per-term GPU/CPU parity oracle. It never tried a tabulated LHY on a device — the scalar path it does cover is exactly the one that was already right.

New gates:

- `test/hamiltonian/test_tabulated_lhy_propagator_parity.jl` — CPU, in the default tiers, so it runs without a GPU
- `test/gpu/test_gpu_tabulated_lhy_parity.jl` — the device leg

Both assert the paths agree **and that a tabulated LHY changes the state at all**. Agreeing on zero is precisely how this hid, so a pure parity check would have passed on the broken code.

**Canary-checked**: reintroducing the coefficient collapse turns the CPU gate red in 8 places. (The first attempt at that canary passed — because my `str.replace` had silently matched nothing after the formatter reflowed the line. Asserting the anchor is now how I do these.)

## Not in this PR

Whether the 12 affected runs need redoing is a call for whoever owns them; the configs are listed in the commit message. Nothing here touches stored results.

Fast tier: 149 files, all passed.


anko9801:
Pushed a **second, independent bug** found on the same term while verifying the first: **the tabulated LHY energy was exactly 2.5× too large**, on CPU as well as GPU.

The tables store `V = dε/dn`; `_lhy_energy` reduced them with `n·V(n)`. Every one of these tables has `ε ∝ n^(5/2)`, for which `n·V = (5/2)ε`.

```
uniform cloud, closed-form  E = ∫ε dV :  3093.31
                _lhy_energy(Tabulated) :  7733.29    ratio 2.500002
```

The tables themselves were never wrong — `∫₀ⁿ V dn'` reproduces the analytic ε to 6.8e-5 — and **the propagator was unaffected**, since it applies `V` directly. So the dynamics were right and only the number printed beside them was inflated. That is why it survived: nothing ever looked unstable.

Affects the same seven modes as the first bug, plus every CPU run, not just GPU.

## The fix, and a second-order point it exposed

`ε(n)` is now the **exact** integral of the same piecewise-linear `V` the propagator evaluates:

$$\varepsilon = \mathrm{cum}[i] + V(x_i)\,\delta + \tfrac12\,\text{slope}\cdot\delta^2, \qquad \delta = n - x_i$$

My first attempt linearly interpolated the cumulative table instead. That makes `dE/dn` the interval **average** of V rather than `V(n)` — a 0.5% mismatch at `n_points = 4000`. Small, but it is the two faces of the same term disagreeing by a discretisation artifact, which is the thing the HamTerm protocol exists to prevent. The FD test caught it, and now `dE/dn == V` holds by construction.

## Gate

`test_lhy_energy_convention.jl` checks the tabulated energy against the analytic closed forms (polar and FM, three densities each), against the **scalar branch — which was always correct**, so this is an independent statement rather than a self-check, and that `dE/dn` recovers `V`. It also pins the specific wrong answer, so a 2.5× cannot come back quietly.

**Canary-checked**: reintroducing `n·V(n)` turns the gate red in 13 places.

Fast tier: 150 files, all passed.

---

Two bugs on one term in one sitting, both invisible to the existing suite, and they fail in opposite directions — the propagator dropped the LHY while the energy inflated it. Worth noting for review: the per-term oracles cover `apply_operator!` ↔ FD-of-`energy_contribution` for the registry HamTerms, but `_lhy_energy` for tabulated modes sat outside that loop.

anko9801:
Merged `main` and closed the gap that merge opened.

**The conflict was not incidental.** #124 landed `SpatialLHY` into the same LHY branch this PR rewrites, and adding both sides is not enough: `SpatialLHY <: AbstractLHY` **directly**, not `<: TabulatedLHY`. So `gpu_lhy_field.jl` — written for `TabulatedLHY` — does not cover it, and dispatch falls through *silently* to the generic CPU broadcast, which closes over host `Vector`s from a device kernel. That is the same failure this PR exists to fix, one level up the type tree.

It also reads the local polarisation, so it arrives at the four-argument `_lhy_potential_field(lhy, density, p, RT)` — a second method again.

**Resolution**

- CPU: `_lhy_needs_spin` is a compile-time trait, so all four call sites branch correctly — plain diagonal step and `_diagonal_step_with_ls!`, each fused and broadcast. That part of the merge is sound.
- GPU: new device method for `SpatialLHY`. It **scans** rather than indexes. The polarisation nodes are the centres of the OCCUPIED bins (`_bin_local_spinors` drops empty ones), so unlike every density table they are **not uniformly spaced** and the O(1) arithmetic this PR uses above it would be silently mis-indexing — the same class of bug again. Reproducing `searchsortedlast` exactly is also what keeps GPU and CPU bit-identical instead of agreeing to a tolerance.

**Gates**

- `test_tabulated_lhy_propagator_parity.jl` — CPU, default tiers: fused == broadcast for `SpatialLHY`, on the plain step AND the light-shift step, real and imaginary time. Neither was covered before: #124's own tests predate `_lhy_field_for_broadcast` and never take the `view` path.
- `test_gpu_tabulated_lhy_parity.jl` — device leg. **17/17 on a local device**, GPU == CPU to 1e-13.

Both also assert the term changes the state at all; agreeing on zero is how the original bug hid.

**Canary-checked**: dropping the `fim` term from `_polarisation_field` turns the new testset red in exactly 4 places (both broadcast call sites × real/imaginary time) and nothing else. The anchor was asserted absent *before* the run — a canary that silently matches nothing passes for the wrong reason.

Fast tier: 151 files, all passed. `_COST` updated for the parity gate (measured 22.2s, was defaulting to 3.0s).

anko9801:
Pushed a **third** bug on the same term. The LBFGS gradient was **exactly zero** for every tabulated LHY.

`_grad_lhy!` read `ws.interactions.c_lhy` only, so any table produced no gradient at all — and unlike the c2/c4/tensor gaps, it did not even warn.

```
PolarContactLHY, vs a finite difference of _lhy_energy:
  |grad_code| = 0.0        |grad_fd| = 628.9
```

Worse than a wrong number: **ITP goes through the propagator and had the LHY; LBFGS goes through the gradient and did not.** The two solvers were minimising different Hamiltonians, and LBFGS reported convergence on the one missing the term.

Fixed rather than warned, because the gradient is available in closed form. `E = ∫ε(n)dV` with `V = dε/dn`, so `δE/δψ̄ = V_LHY(n)·ψ` — the same `_lhy_V` the propagator applies. It reproduces the FD to 1.5e-7. **One function now stands behind all three faces** (propagator, energy, gradient), which is the arrangement the HamTerm protocol is supposed to produce.

Gate: `test_lhy_gradient_all_modes.jl` — FD-vs-gradient across polar_contact / fm_contact / icosahedral / scalar, equality with `V_LHY(n)·ψ`, inactive modes contributing nothing, and the old failure (norm exactly 0) pinned. Canary-checked: 13 failures on restoring the scalar-only read.

---

## Three bugs, one term, one sitting

| face | failure |
|---|---|
| broadcast propagator | dropped the LHY entirely (GPU always took this path) |
| energy | inflated it by exactly 2.5× (CPU and GPU) |
| LBFGS gradient | zeroed it (silently, no warning) |

**They fail in three different directions**, which is why no single symptom ever showed up: the dynamics looked right while the energy was inflated, and the two ground-state solvers disagreed without either complaining.

Each was invisible because each lived on a face the others' tests did not touch. The per-term oracles do enforce `apply_operator!` ↔ FD-of-`energy_contribution` for registry HamTerms — but tabulated LHY reaches the propagator through `_lhy_V`, the energy through `_lhy_energy`, and the gradient through `_grad_lhy!`, and none of those three were tied to each other. Worth considering whether the oracle should cover LHY *modes* and not just the `LHYTerm` wrapper.

Fast tier: 151 files, all passed.

PR #128, opened 2026-07-28
perf(ci): re-measure the parallel-balance cost model — 25 s off the CI critical path

anko9801:
## Why

The LPT bin-packer in `test/runtests.jl` balances test files across the runner's cores by `_COST`. The table had **26 entries against ~250 test files**, and had drifted: the per-push CI's two heaviest files were carrying the 3.0 s default while actually costing **91.5 s** and **84.2 s**.

Under-estimates are the only kind that hurt — the balancer packs a heavy file onto an already-full chunk, and the *makespan* (what the job actually waits on) runs long.

## Measurement

Median per-file time over **four green CI runs** on the GitHub 4-vCPU runner, for both per-push tiers, every file above 6 s. Simulating the LPT packing against each of those four runs individually:

| tier | before | after | perfect-balance floor |
|---|---|---|---|
| `fast` (critical path) | 292.8 s | **267.6 s** | 260.9 s |
| `oracles` | 187.4 s | **160.9 s** | 154.9 s |

Both improve on **every one** of the four runs, not only in median. The new entries land within ~7 s of the theoretical floor, so the balancer is no longer the bottleneck.

Sample of the drift corrected:

| file | was | measured |
|---|---|---|
| `oracles/test_propagator_references.jl` | 11.0 | 91.5 |
| `oracles/test_master_oracle.jl` | 11.0 | 84.2 |
| `test_reference_rhs.jl` | 7.5 | 59.4 |
| `oracles/test_hamiltonian_sign_oracles.jl` | *(default 3.0)* | 52.0 |
| `workflow/test_autopilot.jl` | 12.0 | 49.1 |
| `hamiltonian/test_ddi.jl` | *(default 3.0)* | 33.6 |

## Also

`test_quality.jl` dropped its unconditional `using JET`. The typo check behind it is default-skipped (`SPINORBEC_JET`), so the load — **0.72 s min warm, measured** — bought nothing on every run. JET now loads only when opted in.

## Verification

- `SPINORBEC_TEST_TIER=fast` — 149 files, **all passed**.
- Tier-membership meta-test green (it enforces that every `_COST` key names a real file).
- Both files clean under JuliaFormatter 2.5.0.

Tier lists are untouched — this changes only how files are packed into chunks.

### DIFF
--- a/test/_tiers.jl
+++ b/test/_tiers.jl
@@ -457,9 +457,9 @@ const MANUAL_TESTS_ALLOWLIST = [
 const ORACLE_TESTS = filter(t -> startswith(t, "oracles/"), vcat(FAST_TESTS, CI_EXTRA, FULL_EXTRA))
 
 # ── Parallel-balance cost model ───────────────────────────────────────
-# Per-file cost estimate (seconds), measured on the full tier, used only to
-# balance the parallel chunks (LPT bin-packing in runtests.jl). Only the heavy
-# outliers need an entry; the long tail defaults to _DEFAULT_COST. A wrong
+# Per-file cost estimate (seconds) on the GitHub 4-vCPU runner, used only to
+# balance the parallel chunks (LPT bin-packing in runtests.jl). Only files above
+# ~6 s need an entry; the long tail defaults to _DEFAULT_COST. A wrong
 # estimate costs balance, never correctness — but a silently-stale estimate lets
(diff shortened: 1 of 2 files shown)


anko9801:
Superseded by #129 — same first commit with its message corrected (the '25 s' figure came from an offline simulation that CI did not confirm), plus the on-demand queue that actually removes the imbalance.

PR #129, opened 2026-07-28
perf(ci): hand test files to workers on demand instead of pre-assigning chunks

anko9801:
Supersedes #128 (same first commit, message corrected — see below).

## The problem

The parallel test runner LPT-bin-packed the file list into one **fixed chunk per worker**. That is optimal only if the per-file cost estimates are right, and measured on CI they cannot be: the same file's time swings **±30 % run to run** (runner speed, cache state, JIT order). Over four green CI runs the makespan sat **8–21 % above the perfect-balance floor** — one worker still grinding while the others had nothing left they were *allowed* to touch.

## What I tried first, and what it actually bought

Commit 1 re-fits `_COST` from measured CI medians (the table had 26 entries against ~250 files; the two heaviest per-push files were carrying the 3.0 s default while actually costing **91.5 s** and **84.2 s**).

Measured on CI as makespan / perfect-balance floor — that ratio normalises out runner-speed variance, which is ±40 % between runs:

| tier | before | after |
|---|---|---|
| `oracles` | 1.212 | **1.143** (~10 s, below all four before-runs) |
| `fast` (critical path) | 1.091 | 1.084 (~1.5 s — **inside run-to-run noise**) |

An offline simulation had predicted ~25 s on both. It was wrong: it could not reproduce Julia's unstable `sort` tie-breaking across the ~90 files sharing the 3.0 default, so it packed different chunks than the runner did. #128's title claimed that number; this PR carries the corrected message.

The estimate was never the real problem. **Pre-assignment was.**

## The fix

Commit 2: the workers pull. The parent writes the file list (heaviest first) to a temp dir and starts N identical workers; each takes a file by atomically creating `claim_<i>` with `O_CREAT|O_EXCL` and moves on when it loses the race. The kernel does the mutual exclusion — no lock file, no IPC, no parent-side scheduler. `_COST` survives only as the hand-out order, where being wrong is cheap.

Measured, same machine, same 149 files, `tier=fast`:

| | per-worker totals | makespan / ideal |
|---|---|---|
| static | 432.5 / 426.2 / 396.8 / 456.7 s | 1.067 |
| on-demand | 255.1 / 255.1 / 258.2 / 262.0 s | **1.017** |

On CI's four-worker runner the recovered 7–19 % lands on the critical path.

## Safety

The one failure mode an on-demand queue has that static assignment does not is a worker dying **between** claiming a file and running it — the file would simply not run, and the suite would go green having skipped a test. So a worker writes `done_<i>` only *after* the file returns, and the parent errors loudly on any index with no `done_` marker.

## Verification

- `tier=fast`, 149 files, **all passed**; every file claimed exactly once (`claims == done == n`).
- 3-worker concurrency test on a 6-file queue: each file runs exactly once, all workers exit 0.
- Tier-membership meta-test green (enforces every `_COST` key names a real file).
- Clean under JuliaFormatter 2.5.0.

Tier lists are untouched.


anko9801:
## CI has now confirmed it

Run [30325403549](https://github.com/KozumaLaboratory/BEC-simulation/actions/runs/30325403549) — all jobs green. Per-worker totals from the log, scored as makespan / perfect-balance floor (the ratio that normalises out runner-speed variance):

| tier | static (4 green runs) | static + refit `_COST` (#128) | **on-demand (this PR)** |
|---|---|---|---|
| `fast` (critical path) | 1.080 / 1.062 / 1.130 / 1.101 → **median 1.084** | 1.084 | **1.006** |
| `oracles` | 1.203 / 1.244 / 1.194 / 1.221 → **median 1.203** | 1.143 | **1.020** |

Worker totals on this run — `fast`: 277.6 / 277.4 / 280.5 / 279.4 s; `oracles`: 159.9 / 166.0 / 160.7 / 164.1 s.

At this run's scale that is **~22 s recovered on the critical path** and ~30 s on `oracles`. 1.006 is essentially the floor — the residual is one file's granularity, so there is nothing further to win from scheduling.

## What is left, measured, and deliberately not done here

- **`needs: changes` costs ~24 s of serial prefix.** Run created 03:14:58, the `changes` job ran 03:15:07→03:15:19, `test` could not start until 03:15:22. Moving the docs-only detection in-job would recover ~15 s, but it would rework the branch-protection fix from e283733f and the dangerous failure direction (a code PR that skips its tests) is not worth ~15 s without a careful design. Flagging, not doing.
- **`@compile_workload` costs ~8 s** of SpinorBEC precompile, paid in both jobs on every PR (paired interleaved A/B, MIN statistic: 33.5 s on vs 25.6 s off; an earlier unpaired reading said 22 s and was contention-inflated). `LocalPreferences.toml` `[SpinorBEC] precompile_workload = false` does work, but ~8 s did not justify making the CI image differ from what ships.
- **Repo Actions cache is at 10.44 GB against GitHub's 10 GB ceiling** (22 entries). Currently harmless — all jobs share one key so the oldest `run_id` evicts first, and restores are hitting (486 MB in ~7 s, "158 already precompiled"). Worth watching: it is the same setup as the 2026-06-21 incident.

PR #134, opened 2026-07-28
feat(lhy): wire `spatial` into YAML, and make n_max/n_points actually do something

anko9801:
Rebuilt on `main` after #132 merged (supersedes the auto-closed #133 — deleting
its base branch closed it rather than retargeting).

Two things in the same block, because they are the same wiring.

## `lhy: {n_max, n_points}` were read by nothing

Both have been in `LHY_SCHEMA` since the C6 block landed. `_resolve_lhy_block!`
normalised only `kind` and `c_lhy`; `_build_spinor_lhy` hard-coded
`n_max = 3 × max|ψ|²` and let `n_points` fall through to each builder's own
default. A user writing `n_points: 4000` got 200, `n_max: 11` got whatever the
cloud implied, and nothing said so — a YAML key that reads as live and is not.
**Any config predating this ran at the defaults regardless of what it asked for.**

They now ride in one concrete `LHYTableOpts(n_max, n_points, n_bins)` from the
resolver through `find_ground_state` to `make_workspace`. One struct rather than
three kwargs deliberately: `make_workspace` is the inference hot path, and the
Dict lookups producing it are narrowed with `::LHYTableOpts` at the pipeline
boundary so no `Any` reaches the Workspace path.

## `kind: spatial`

`SpatialLHY` landed in #124 with a builder and no way to ask for it. Now
`lhy: {kind: spatial, n_bins: N}` tabulates `e₁(p)` from the actual local
spinors, at one BdG solve per **occupied** `|⟨F⟩|/F` bin — so the cost is
independent of grid size, and `n_bins` is a cost knob.

`compute_spatial_lhy` returns `nothing` for a cloud uniform enough that a
single-spinor table is already exact. That is a correct answer, but `nothing`
would hit make_workspace's silent-zero throw, so the builder falls back to
`:full_bdg` — the single-spinor table that IS right in that regime — rather than
erroring or, worse, running with no LHY. Same for `psi_init === nothing`.

Expect `full_bdg`'s dynamic-instability warning on real textures, and do not
read it as a defect: a bin's representative spinor is lifted out of the cloud
and is not a solution of the *uniform* mean-field problem at its own density, so
the instability is a property of that fictitious uniform system. Documented at
the call site rather than suppressed.

Depends on #132 (already merged): routing the LBFGS gradient through `_lhy_V`
gave `SpatialLHY` a real gradient. Without it, making this kind reachable from
YAML would MethodError in LBFGS.

## Verification

- `test/workflow/test_lhy_block_wiring.jl` — new, 55 assertions, fast tier.
  Includes a **per-mode** check that all six `n`-tabulated builders honour the
  knobs: the plumbing is seven separate call sites and one missed edit is the
  exact shape of the bug being fixed. Also gates the uniform-cloud fallback, the
  resolver's defaults, and that `LHYTableOpts` is fully concrete.
- End-to-end smoke: `make_workspace(spinor_lhy=:spatial)` → `SpatialLHY` in the
  workspace → both `energy_contribution` and `apply_operator!` finite and
  non-zero.
- fast tier on the pre-rebuild branch: **153 files, all passed**; the four
  directly affected suites re-run green after the rebuild.
- `test_lhy_polar.jl` was the one caller of `_build_spinor_lhy` outside
  `make_workspace` and needed the new positional; fixed here.


anko9801:
Merged current `main` in (which now has #136, the mode×face gate) and re-verified. That combination was untested by either PR's CI — #136 ran before `spatial` was in the enum, #134 ran before the gate existed — and it did surface something.

The gate's forward-compat `spatial` fixture returned `nothing`. Cause: adjacent components stay spin-coherent, so at F=6 an (m=6, m=5) mixture holds |⟨F⟩|/F ≈ 0.92 across the whole cloud — one occupied bin, spread 0. That is `compute_spatial_lhy` behaving correctly (no texture in the axis it tabulates against), so it declines to build. Fixture now mixes m=+F with m=0, matching `test_spatial_lhy.jl`.

Gate is 94 → 105 assertions, all 10 modes. oracles 63 files and fast 154 files both all-passed on the merged state.

PR #145, opened 2026-07-28
perf(rtp): round 9 — per-voxel device-sized rotation + fused V half-step (2.84× on the step)

anko9801:
> **Stacked on #142** — base is `perf/rtp-shared-spin-rotation`, so this diff is exactly the six commits of the next round. Merge #142 first.

Round 9 of the RTP hot path. Headline on RTX 5070 Ti, Eu-151 F=6 (D=13) F64 + DDI at 64³: **87.2 → 30.8 ms/step (2.84×)**, with GPU-busy share going 34.3% → 82.4% of the trace.

## 1. Two FP64 divisions out of the Horner body

`bench/micro_spin_taylor.jl` sweeps the Horner degree and shows the kernel is purely compute-bound — time exactly linear in K with a zero memory intercept — so every operation in the loop is on the critical path. The body carried `z / T(k)`, a Complex/Real divide, i.e. **two FP64 divisions per degree per lane** on hardware where FP64 divide is the most expensive op available.

`z` is purely imaginary in real time and purely real in imaginary time, so the coefficient is really the real number `scale/k`. Tabulated once per call, with the multiply by ∓i folded into the body as a component swap. Values are bit-identical to the old expression (verified exhaustively over random inputs).

| K | before | after | |
|---|---|---|---|
| 4 | 3.367 ms | 1.066 ms | 3.16× |
| 8 | 6.798 ms | 1.690 ms | 4.02× |
| 12 | 9.503 ms | 3.524 ms | 2.70× |
| 16 | 13.480 ms | 3.801 ms | 3.55× |

## 2. Size the rotation per voxel, on the device

Two independent costs with one cause: the Horner degree was decided **on the host** from `R = max_r|v(r)|`.

- That read a device reduction back to the host, draining the stream once per rotation — 12× per RTP step. The GPU sat idle for two thirds of the step.
- It was also the wrong degree almost everywhere: a trapped cloud is mostly dilute tail where 2–3 terms suffice, but every voxel paid the peak's degree.

Each voxel now sizes itself: halve the angle until `R ≤ _SPIN_TAYLOR_RSAFE[]`, apply the halved rotation 2^s times, take the smallest degree meeting the same `R^(k+1)/(k+1)! ≤ tol` bound the host used. Degree and repeat count are raised to the **warp** maximum first, since `_shfl_c` carries a full-warp mask and the two voxels sharing a warp must agree on the trip count. The backward-error contract is unchanged and now holds *per voxel* rather than in the max.

The halving makes the Taylor path exact at **any** R, so the "R too large" Euler fallback is gone — `_SPIN_TAYLOR_RMAX`, `_SPIN_TAYLOR_R_OVERRIDE`, `_spin_rotation_R` and `_taylor_degree` are deleted with their callers migrated. Exact Euler stays as the D > 16 path and the parity reference.

| | before | after |
|---|---|---|
| step | 87.245 ms | 30.765 ms (2.84× cumulative) |
| GPU busy | 57.67 ms | 26.78 ms |
| GPU busy % | 34.3 | 82.4 |
| rotation (12 calls) | 45.73 ms | 17.10 ms |

## 3. The whole V half-step as one pass over ψ

In the Eu production shape the half-step reduces to `diag(t/2)·SM(t/2)·DDI(t)·SM(t/2)·diag(t/2)`, whose middle three factors are the *same* operator $\exp(-i a\,(v\cdot F))$ with different vector fields. Applied one at a time that is eight passes of ψ through HBM, three recomputations of ⟨F⟩, and five launches — for arithmetic that fits in one lane's registers.

Three facts collapse it to one pass:

- the midpoint predictor-corrector **freezes** the mean field over the half-step, so all three rotations want the same ⟨F⟩ — and the DDI convolution already computes it on its way to Φ;
- a spin rotation is unitary, so it cannot move the total density; with `psi_mf` frozen both diagonal halves carry the identical voxel phase, computed once;
- only the per-component Zeeman factor differs between the halves, and that is D numbers, not a field.

Per half-step: **8 ψ passes → 3, 3 spin-density passes → 1, 5 launches → 2.**

**This is not a different splitting.** Same order, same operators, same arithmetic — only the round-trips are gone, so the result is bit-identical and `test/oracles/test_spin_chain_fusion_parity.jl` demands exactly that rather than a tolerance. It earned its keep immediately by catching a 1-ulp multiply association (`psi*vph*zph` vs the diagonal kernel's `psi*(vph*zph)`) that any tolerance would have waved through. Canaried: dropping one rotation from the fused kernel moves the state by 3.5e-2.

`_spin_chain_reason` is the single list of everything the fused path would otherwise silently drop, with one test arm per entry.

```
bench/rtp_fusion_ab.jl 64 10 — RTX 5070 Ti, Eu151 F=6 (D=13) F64 + DDI
  operator-by-operator   36.189 ms/step
  fused                  30.911 ms/step   1.171x
```

## Two rejected attempts, kept with their measurements

- **Shared-memory staging of ψ** (`SPIN_STAGE_SHARED[]`, off by default). The coalescing argument is real — on an H100 the same kernel is memory-bound (time *flat* in the degree, `t(16)/t(4) = 1.00`) and reaches only 894 GB/s of ~3.35 TB/s peak. But on the WSL2 RTX 5070 Ti the kernel is compute-bound (1/64-rate FP64) and staging is 5–15% *slower*, at 48 registers with no spill. Shipped behind a toggle with `bench/rtp_stage_ab.jl` as the per-machine A/B, rather than picked blind.
- **Batching the DDI convolution's 6 rFFTs** — worth ~3%, for a `DDIBuffers` field-type change that reaches the `Workspace` type parameters. Not worth it.

Both are recorded in `bench/perf_targets.txt` with the caveat that everything here was measured on a 1/64-rate-FP64 WSL2 card, so **the ranking does not transfer to an H100**.

## Two measurement traps, both hit on the way

Worth repeating because they invert conclusions:

- A cross-run comparison of two `profile_rtp.jl` invocations showed the fusion as a **4% regression** — process-to-process clock and JIT drift is larger than the effect.
- Without reclaiming the pool between arms, one fused arm read 12.3 ms against 30.4 ms for the same arm moments later — a spurious **2.96×**.

Interleaved arms, reclaimed pool, min of 3, or the number means nothing.

## A latent parallel-mode bug this surfaced

`test/oracles/test_harmonic_virial.jl` built ψ as a host `Array` but asked `make_workspace` for no particular backend. `_resolve_backend` auto-selects CUDA for a 3D grid ≥ 24³ whenever `cuda_functional()` — true only if some **earlier file in the same parallel-worker session** happened to `import CUDA`. So `ws.potential_values` came back a `CuArray`, the host ψ broadcast scalar-indexed, and the gate errored.

It passed standalone and failed under `SPINORBEC_TEST_WORKERS` — the signature of exactly the cross-file coupling CLAUDE.md warns about for parallel mode. It went red now only because adding `test_spin_chain_fusion_parity.jl` (which imports CUDA) to the ci tier changed which files share a worker session. Pinned to the CPU backend.

---

**Provenance:** these six commits were authored in a separate session and were sitting unpushed on a local worktree; I rebased them onto #142 and surfaced them. The measurements above are quoted from their commit messages — I have **not** independently re-run the benchmarks. CI status is reported by the checks below.


anko9801:
Superseded by #148, which contains all six commits here plus two more: a revert of the shared-memory staging toggle this PR shipped, and a Taylor-tolerance tightening to 1e-13 that per-voxel degrees made binding.

That branch also carries H100 measurements (2.1×), where the numbers here were taken on a 1/64-rate-FP64 WSL2 card — and the commit messages here explicitly warn the ranking does not transfer. Closing in favour of the better-measured superset; nothing is lost.

Context: these commits were sitting unpushed on a local worktree when I surfaced them, which is why the two PRs exist.

PR #159, opened 2026-07-28
research(lhy): the LHY-insufficiency conclusion holds; its evidence was vacuous

anko9801:
First re-derivation from the #150 audit list. **The conclusion survives. The comparison it rested on did not** — and it took three coefficient fixes before that comparison became a real test.

*(This PR was drafted twice with the wrong answer; see "Two false starts" below. Both were caught before merge.)*

## What the note claimed

`eu_collapse_lhy_insufficient.md` (2026-05-07) compared five LHY treatments on the Eu F=6 EdH post-quench collapse, found identical density profiles, and concluded **"LHY is sub-leading vs the mean-field DDI attraction."**

## Why its evidence was vacuous

That config is `backend: gpu`. Until #125 every *tabulated* LHY was silently collapsed to `c_lhy = 0` on the GPU broadcast path:

- `polar_contact`, `polar_dipolar`, `full_bdg` → **no LHY at all**
- `scalar` → coefficient short by $\pi(a_s/a_{ho})\sqrt N$, **2.34×** at this $N=10^4$ (#108)
- `off` → LHY-off by construction

"All five identical" was four flavours of *off* plus one weak one. **The observation was guaranteed regardless of the physics.**

## Re-run on current `main`

Which also carries #158 (closed-form tables $N_{atoms}$ too large):

| mode | $E$ | peak $n$ | FWHM$_z$ | $M_z$ |
|---|---:|---:|---:|---:|
| off | −881.086 | 0.01014 | 10 | −5.4104 |
| scalar | −880.993 | 0.00968 | 10 | −5.4567 |
| polar_contact | −881.022 | 0.00982 | 10 | −5.4431 |
| fm_contact | −881.022 | 0.00982 | 10 | −5.4431 |
| *full_bdg* | *−847.324* | *0.00054* | *22* | *−5.9955* |

The four usable modes agree to **4.7 %** in peak density, and the closed forms are **active** — all three differ from each other and from `off`, they are simply small in this regime. So the conclusion now rests on a comparison that *could* have come out otherwise, which is what makes it evidence.

ablation

`full_bdg` is excluded on its own report:

```
Warning: FullBdG LHY: mean field is dynamically unstable (max Im ω = 213.0).
  ... so ε_LHY is scheme-dependent here. This is a property of the state.
    E=-847.3236 conv=false
```

## Two false starts, recorded in the audit doc

Both mine, both caught by checking the code's own report rather than by reasoning:

1. The first re-run reported the closed forms cutting peak density **15×** and arresting the collapse. That was **#158, which merged 22 minutes after the run**. Caught with `git merge-base --is-ancestor` — by asking whether the run contained the fix instead of assuming it did.
2. The surviving 19× outlier was nearly quoted as "LHY can arrest this collapse". It was `full_bdg`, which prints a warning saying its $\epsilon_{LHY}$ is scheme-dependent for exactly this state, and did not converge.

## Generalisation for the rest of the audit list

**A stale null result is the dangerous kind**, because a broken knob and a knob that does not matter look identical. Check the run's commit against the fix list, and read the run's own warnings, before comparing numbers.

## Data hygiene

Run output lives off-repo on TSUBAME: `/gs/bs/work/7/uk07267/spinorbec-runs/lhy_ablation_v2` (862 MB). The first attempt wrote 866 MB into the repo worktree; that was moved out.

Worth noting for the next person: `run_yaml` does **not** read `SPINORBEC_STORE` — that env var covers the CAS store and the GS stage cache only. `run_yaml`'s output root is its `base_dir` argument.


anko9801:
⚠️ **Hold — this PR's own numbers are stale, by 22 minutes.**

The ablation ran at `2026-07-28T18:58Z`. #158 (*"the closed-form spinor tables were exactly N_atoms too large"*) merged at `2026-07-28T19:20Z`. Verified: `git merge-base --is-ancestor 87c0ae5d <this branch>` → **not an ancestor**, so the run did not contain that fix.

That fix targets exactly the three modes this PR's conclusion rests on — `polar_contact`, `fm_contact`, `full_bdg`. They ran with the closed-form table $N_{atoms} = 10^4$ too large, so "the F-aware closed forms arrest the collapse" is measured with LHY absurdly too strong and cannot stand as written.

Re-running on current `main`. The framing may well survive — the *original* note's three closed-form rows still ran with `c_lhy = 0` on the GPU, which is independent of this — but the magnitudes in the table here are void until re-measured.

Converting to draft rather than merging.

PR #162, opened 2026-07-28
feat(barnett): fig_m6_spatial in time, and a vortex indicator the m-panels cannot give

anko9801:
`fig_m6_spatial` shows per-$m$ column densities, and you cannot tell from it whether there is a vortex. The rings in those panels are **not** vortices: a spin texture whose direction varies with radius gives *every* $m$ component a ring through the Wigner-$d$ weighting, flow or no flow. So the question the figure is asked — is there a vortex, and in which arm — is not one it can answer.

## What can answer it

A vortex is a hole in the **total** density with quantised circulation of the **total** current,

$$\mathbf{j} = \sum_c \mathrm{Im}(\psi_c^* \nabla \psi_c), \qquad \mathbf{v} = \mathbf{j}/n, \qquad w = \frac{1}{2\pi}\oint \mathbf{v}\cdot d\boldsymbol{\ell}$$

That is invariant under a uniform spin rotation, so the component rings contribute **exactly nothing** to it while a real core still shows.

## The result

At $t = 47.2$ all three arms show rings in every $m$ panel. Only $\ell = -1$ carries a vortex marker. Over the full run:

| $t$ | $\ell=+1$ | $\ell=0$ | $\ell=-1$ |
|---|---|---|---|
| 0.4 | **+1** ($L_z$ +0.96) | 0 ($L_z$ −0.04) | **−1** ($L_z$ −1.03) |
| 4.0 | 0 ($L_z$ **−0.94**) | −1 | **−1** |
| 16 | 0 | 0 | **−1** |
| 50 | 0 | 0 | **−1** ($L_z$ −2.25) |

The $\ell=-1$ vortex survives all 50 time units; $\ell=+1$ loses its vortex by $t \approx 4$ and its $L_z$ changes sign. That is the floor-rectification asymmetry `run_m6_imprint.jl` was built to test, now visible rather than inferred.

## Two things the positive control caught

The imprint is its own control — $\ell = \pm 1$ is a known vortex, $\ell = 0$ a known non-vortex — and it failed twice before passing:

- **The loop must be $R=2$ cells across, not one plaquette.** $v \sim 1/r$ at a core, so a single-plaquette circulation returns $w = 0.238$ on an exact $f(r)e^{i\varphi}$ vortex instead of 1. Measured: $R{=}2 \to 0.982$, $R{=}3 \to 0.990$, $R{=}5 \to 0.994$.
- **The density gate must be a neighbourhood, not the four corners.** A core is precisely where the density vanishes; with $f(r)\sim r$ the corners sit at ~0.4 % of peak, so a corner gate rejects every vortex it is meant to find.

Raw $w$ is stored per defect, not just the rounded charge, so "how quantised is quantised" stays a question the reader can re-ask without re-running the physics. `mass_vortex_counts_strict` ($|w-n| < 0.25$) is the number the movie leads with.

## Why a re-run rather than a re-read

`run_m6_imprint.jl` streams $\psi$ to node-local scratch and pulls exactly one time out of it; that scratch is wiped at job end. `run_m6_movie.jl` runs the same imprint + quench and reduces **every** frame inside the job, so only ~29 MB/arm leaves it.

Three storage limits, all hit for real, all now handled:
- group `/gs/fs` quota is 1 TB and was **at** it (1000.06/1000.00 GB) → run dirs go to the personal work area via `SPINORBEC_STORE`. Note `run_yaml` takes this as a `base_dir` **kwarg**; it does not read the environment itself.
- per-job `/tmp` on `gpu_1` is small — 250 streamed 80×80×40 frames (~6.6 GB) SIGBUSed the snapshot mmap mid-dynamics, twice → `RB_SAVE_EVERY >= 1000`.
- JLD2 mmap on Lustre SIGBUSes → archives are built on `/tmp` and copied.

`RB_TAG` is carried in the config spec on purpose: the content-addressed run key is the spec alone and does **not** include the source version, so a byte-identical config silently reuses a result computed by pre-bugfix code.

## Merge notes

Brings in #142 (spin-mixing shares the DDI rotation kernel — 80 % of the RTP step) and #147 (`orbital_angular_momentum` plan cache, hit once per frame here). `COMBINED_SPIN_STEP_ENABLED[]` is deliberately **not** enabled: a different splitting, equal in order but differing at $O(dt^3)$, so it would move the numbers the existing m6 figures were computed with.

⚠️ `src/hamiltonian/coefficients.jl` now holds **two** declarations of the scalar LHY coefficient that disagree — 1.98 vs 0.076 on the same input. `scalar_lhy_coefficient` is main's PR #108 fix; `compute_gamma_lhy` is reached only by the rotating_basis ground state, whose own test pins the current values. I flagged it with a TODO rather than unifying, because unifying moves that path's results and that is not my call. **This should not stay.**

Tests: fast tier green (the tier-membership meta-test caught the new file as an orphan and it is now registered); `test_vortex_density_movie.jl` 43/43, including the canary that separates the two traces — a state whose components differ only by a real, node-free envelope has zero mass current and must report zero vortices, while the per-component winding fires.

Deliverable: `runs/eu_barnett_rotfield_clean/figures/fig_m6_spatial_movie.mp4` (5.0 s, 150 frames, 30 fps).


anko9801:
Closing as superseded — checked what would be lost first, since this is 47 commits of work.

**Every code file here has already landed on `main` by another route.** Diffing the branch against `main`, the set of files this branch has that `main` lacks contains **no `src/`, `test/` or `scripts/` entries at all**. Specifically:

- `vortex_density_movie.jl` and `viz_dynamics_movie.py` → `main` via 96a75d87 *"watch an excitation happen — vortices, density, spin, as a movie"* and 15296ea5 *"define the vortex on the TOTAL density — mass-current circulation"*, which is exactly what this PR's title claims ("fig_m6_spatial in time, and a vortex indicator the m-panels cannot give").
- `run_m6_movie.jl`, `tsubame_m6_movie.sh`, the `edh_*.yaml` set → `main` via #169, the archive of the campaign orphaned when #62 was closed.

What is unique to this branch is **17 output artifacts** — 14 PDFs and 3 MP4s under `runs/eu_barnett_rotfield_clean/`. Two reasons not to carry them over:

1. They are products of the round #141 supersedes, whose **±Ω arms were not mirror images**: `SinusoidalWaveform` is *sin*, so `phase_x = ±π/2` gave B(0) = ±x̂ and the +Ω arm started 180° anti-aligned with the ground-state spin. The rotation senses were right, but the arms differed by an initial condition too, so any chirality magnitude read off `barnett_direction*.pdf` is confounded.
2. The branch also still carries `Bz: "-0.01 Gauss"` in the five `edh_*.yaml`, which #170 corrected on `main` — the ground state was solved under a field that disfavours its own `m_minus_F` seed, and the dynamics block in the same file already used +0.01. Merging this as-is would silently reintroduce that.

Nothing else is lost. If any of those 17 artifacts is still wanted, it should come back with a note about the arm asymmetry, not as-is.

Issue #172, opened 2026-07-28
LHY term is 97 % of the total energy for tabulated closed forms at Eu F=6 (46 % energy drift), while ScalarLHY reports lhy = +0


anko9801:
Found while re-deriving Figure 2 of `four_figure_spec_2026_05_26.md` (#171). Two separate anomalies in the same four-arm comparison, both on `backend: cpu`, both at (or within one commit of) current `main`.

## Reproducer

`runs/eu_k3_lhy_control/LHY_polar_contact_K0.yaml` and `LHY_icosahedral_K0.yaml`, unmodified. Eu151, F=6, 32³ / box 12, N=30000, `c1_ratio: -0.005`, DDI non-secular, `initial_state: m_minus_F`.

Compared against the same config with `lhy: {kind: scalar}` (shipped) and `lhy: {kind: none}` (I had to add this arm; the directory has no LHY-off config).

## Anomaly 1 — tabulated closed forms dominate the Hamiltonian

| arm | E drift over the run | `lhy` | `total` |
|---|---:|---:|---:|
| none | 2.1e-08 | 0 | +2.24 |
| scalar | 6.8e-08 | 0 | +2.25 |
| polar_contact | **4.54e-01** | **+1664** | +1705 |
| icosahedral | **4.58e-01** | **+1653** | +1694 |

`lhy` is 97 % of the total, against a complete mean field of ≈ +2.2 in the other two arms — three orders of magnitude larger than every other term (`kinetic` +35.3, `trap` +4.9, `density` +2.3, `ddi` −0.16). The dynamics energy series falls 3123 → 1694, i.e. 46 % non-conservation, and the initial peak density is 4× that of the clean arms (0.0185 vs 0.0046).

ITP still reports `converged: true`, so nothing gates this.

## Anomaly 2 — `ScalarLHY` reports zero energy

[...]

PR #174, opened 2026-07-29
fix(lhy): the dynamics phase never resolved its own `lhy:` block

anko9801:
Root cause of the first half of #172, found while re-deriving Figure 2 (#171).

`_resolve_lhy_block!` derives `interactions.c_lhy` for the scalar modes and writes the `lhy_opts` that carries `n_atoms` for the tabulated ones. It runs inside `_resolve_derived_params!` — and the only caller of that is `run_step_ground_state.jl`. So a `dynamics: {lhy: …}` block reached `make_workspace` unresolved, failing in two opposite directions, both silent:

**Tabulated closed forms** got no `lhy_opts`, so the fallback `LHYTableOpts()` supplied `n_atoms = 1`. That factor is the unit conversion, not a table knob — `n` is normalised to $\int |\psi|^2 dV = 1$ while $c_0$ already carries $N$ — so the table came out **exactly `N_atoms` times too strong, in the propagator as well as the energy**. On the Eu F=6 K₃=0 factorial (N=30000):

```
E_LHY = +1664   against a whole mean field of +2.2   (97 % of the total energy)
E drift = 46 %  (3123 -> 1694)
```

Evaluating the *same* energy face on the *same* saved state with a correctly built table gives **+0.0915** — a ratio of exactly `N_atoms`.

**Scalar / quasi_2d** got `interactions.c_lhy = 0`: the dynamics `interactions` dict is re-parsed by `_parse_gs_interactions`, and the GS resolver only ever wrote `c_lhy` onto the *ground_state* block's dict. So the dynamics phase ran with **no LHY at all** while `lhy: {kind: scalar}` sat in the YAML. Being *absent* rather than wrong, it conserved energy to 7e-8 and reported `lhy = +0` — the failure mode that looks exactly like success.

## Fix

`_resolve_dyn_lhy!` runs the same resolver on the dynamics block, **before** `_parse_gs_interactions` (which is what reads the `c_lhy` it writes). One declaration, both phases.

The `c_total` interaction form carries no atom number and so cannot be normalised at all. That returns `false` and the caller `@warn`s, rather than silently keeping `n_atoms = 1`.

## Why every existing gate was green

- **`test_dynamics_lhy_plumbing.jl`** sets `c_lhy` **by hand** — `InteractionParams(Dict(...); c_lhy=2.5)` — then checks `make_workspace` picks it up. It bypassed the exact plumbing that was broken. It covers the schema and `make_workspace`'s silent-zero defence, not whether the pipeline threads a value.
- **`test_lhy_energy_convention.jl`** (17/17) and **`test_tabulated_lhy_propagator_parity.jl`** (42/42) are *internal*-consistency gates (ε vs ∫V dn; propagator paths against each other). A uniformly rescaled table satisfies both, which is why an `N_atoms`-fold error was invisible to them.

The new gate pins **both halves against the ground_state resolution of the same block** — the two phases must normalise identically — and asserts the *consequence* (table strength changes by exactly `N_atoms`) rather than only the field value.

## Scope

Affects every YAML run whose `dynamics:` step carries an `lhy:` block. Type **A** (code correctness).

Closes the first half of #172. The `off` arm's separate 2.3339 → 1.0502 change is **not** LHY — that arm has none — and stays open.

`tier=fast`: 163 files, all passed. `tier=ci` running locally (required checks do not cover it). A 4-arm end-to-end re-run is in flight to confirm the corrected Figure 2 numbers; I'll post them here.

### DIFF
--- a/src/workflow/experiments/pipeline/run_step_dynamics.jl
+++ b/src/workflow/experiments/pipeline/run_step_dynamics.jl
@@ -30,6 +30,49 @@ end
     return TimeDependentInteractions(c0_wf, c1_wf)
 end
 
+"""
+    _resolve_dyn_lhy!(p, atom, c_dd_val) -> Bool
+
+Resolve a `dynamics:` step's own `lhy:` block, in place, exactly as the
+ground_state step does.
+
+`_resolve_lhy_block!` — which derives `interactions.c_lhy` for the scalar modes
+and writes the `lhy_opts` carrying `n_atoms` for the tabulated ones — runs inside
+`_resolve_derived_params!`, and the only caller of that is
+`run_step_ground_state.jl`. So a `dynamics: {lhy: …}` block reached
+`make_workspace` with **neither**:
+
+  * scalar / quasi_2d modes: `interactions.c_lhy` stayed 0, because the dynamics
+    `interactions` dict is re-parsed by `_parse_gs_interactions` and the GS
+    resolver only ever wrote `c_lhy` onto the *ground_state* block's dict. The
+    dynamics phase then ran with **no LHY at all** while `lhy: {kind: scalar}`
+    sat in the YAML — and, being absent rather than wrong, it conserved energy
+    perfectly and reported `lhy = +0`.
+  * tabulated modes: `lhy_opts` was missing, so the fallback
+    `LHYTableOpts()` supplied `n_atoms = 1`. That factor is the unit conversion,
+    not a table knob (see `_resolve_lhy_block!`): `n` is normalised to
+    `∫|ψ|²dV = 1` while `c₀` already carries `N`, so `n_atoms = 1` makes the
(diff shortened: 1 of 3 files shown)


anko9801:
Local `tier=ci`: **257 files across 10 workers, all passed** — run because the required checks do not cover that tier and this touches `workflow/experiments/pipeline/`.

Full gate status: `tier=fast` 163/163, `tier=ci` 257/257, Oracle gates (`tier=oracles`) pass, JuliaFormatter pass.

The 4-arm end-to-end re-run is still going (~26 min per arm on CPU; the `off` arm is reproducing its previous energy +2.24363 exactly, as it must — that arm has no LHY). Corrected Figure 2 numbers will go to #172 and into the figure spec once all four land; they are confirmation of the physics values, not of this fix, which is gated by the unit test above.

PR #183, opened 2026-07-29
perf(itp): 2.02× on H100 32³ — re-fuse the DDI fast paths the padding default-flip bypassed

anko9801:
Makes an ITP step faster. All numbers measured on TSUBAME, baseline and
optimised worktrees on the **same node** per pair (`bench/bench_itp_step.jl`,
¹⁵¹Eu F=6 D=13, DDI padded, `lhy: none`):

| config | before | after | Δ |
|---|---|---|---|
| H100, 32³ | 0.992 ms | **0.490 ms** | −0.50 ms (**2.02×**) |
| H100, 64³ | 3.234 ms | **2.640 ms** | −0.59 ms (−18.4 %) |
| cpu_4, 32³ | 65.84 ms | **60.35 ms** | −5.49 ms (−8.3 %) |
| cpu_4, 48³ | 268.5 ms | **237.9 ms** | −30.6 ms (−11.4 %) |

Host allocation per step: 1.572 → 0.080 MB (CPU 32³), 0.380 → 0.083 MB (GPU).

## What was wrong

`DDI_PADDED_DEFAULT` flipped to `true` this morning (9c117c05, correct — it
fixes a 2-5 % dipolar field error). Padded buffers are *larger than* `n_pts`,
and three hot-path branches key off exactly that test. Every run since the flip
took the slow arm of all three:

1. **The fused GPU spin-density kernel stopped being reached at all.** It was
   gated on `size(fx) == n_pts`; padded fell back to `_spin_density_corner_bcast!`
   — ≈37 kernel launches at D=13, twice per ITP step. Fixed by giving the kernel
   a destination index map instead of a second implementation.
2. **The rotation materialised Φ's corner.** `phi[CartesianIndices(n_pts)]` is
   0.764 MB of host allocation per DDI call at 32³ (exactly 3·n³·8 B), and on
   the GPU three device allocations plus three gather kernels. Replaced by a
   voxel → buffer index map read inside the angle pre-pass and inside the Taylor
   kernel — nothing is copied on either layout now.
3. **`_compute_and_convolve_ddi_padded!` re-zeroed the whole padded volume**
   every call. Those zeros were already there: the only writes are the
   `[1:n_pts...]` corner, which the spin density assigns in full and the Orszag
   F-filter is documented to leave the rest of alone.

Independently of the flip:

4. **The spin-mixing angle pre-pass was serial** — D `abs2`, D−1 complex
   products, an `atan`, an `acos` and a `sqrt` per voxel, single-threaded, while
   the identically-shaped `_ddi_compute_angles!` is threaded. It was serial only
   to carry `max θ` in a loop variable; θ is written to the cache anyway, so the
   guard reads it back (θ ≥ 0, so `max θ² == (max θ)²` — the same number).
   `spin_mixing` was **53 % of a CPU ITP step**: 8.71 → 6.93 ms/call at 32³.

## Shape of the fix

The padded/contiguous split is now one code path, not two arms:
`_voxel_index` (`src/foundation/voxel_index.jl`) — `nothing` compiles to the
bare linear index, a `CartesianIndices(n_pts)` is the padded corner. One
declaration, shared by the CPU loops and the CUDA kernels, so the two devices
cannot drift on it. `_ddi_crop_phi` is deleted.

## Gates

New: `test/hamiltonian/test_ddi_padded_zero_pad_invariant.jl` — dropping the
memset makes "the convolution zero-pad is zero" an invariant rather than a
coincidence, so it needs a gate. It carries a **positive control**: dirtying one
pad voxel DOES move Φ, so the assertion is a real constraint on every writer of
`F_*_pad` and not a tautology.

Green on TSUBAME (`scripts/tsubame/submit_itp_verify.sh`):
`test_ddi_padded_zero_pad_invariant`, `test_ddi_padded`, `test_ddi`,
`test_spin_mixing`, `test_ddi_truncated_kernel`, `test_itp_ddi_strang_save_every`,
`test_ddi_translation_covariance`, `test_spin_density_consistency`.
Full `ci` tier: see the comment below.

## Not done, and why

The CPU **Euler 5-stage rotation core** is now ~63 % of a CPU ITP step
(~6.3 ms/call × 6 calls at 32³). It moves ψ through cache ~18 times — 5 phase
passes plus 4 gemms. The GPU already replaced it with a tridiagonal
Taylor–Horner (`gpu_spin_rotation_taylor.jl`, 2 passes, measured 3.6× at D=13);
there is no CPU counterpart. That is the next large ITP win, but it is **not
bit-identical** — it needs its own backward-error contract and parity gate,
which is a separate change and a separate accuracy discussion.


anko9801:
**Verification (all on TSUBAME).**

`ci` tier, cpu_16, 8 workers, 430 s: **258 files, all passed** (`TEST_RC=0`).
Run via `scripts/tsubame/submit_test_tier_ci.sh` — main's required checks are
fast + oracles + formatter only, so a green PR alone would say nothing about
`ci`, and this touches `hamiltonian/`, `foundation/` and the CUDA ext.

Targeted gates on a CPU node, all green: `test_ddi_padded_zero_pad_invariant`,
`test_ddi_padded`, `test_ddi`, `test_spin_mixing`, `test_ddi_truncated_kernel`,
`test_itp_ddi_strang_save_every`, `test_ddi_translation_covariance`,
`test_spin_density_consistency`.

Unrelated observation from the same run, for whoever owns the test balance:
`validation/test_dipolar_supersolid_tube.jl` took **274.8 s** against a `_COST`
estimate of 3.0 s — 64 % of the whole tier's wall-clock on one file, and the
runner's own drift warning flags it. Not touched here.

GPU gates (`scripts/tsubame/submit_gpu_gates.sh`) are queued on gpu_1; result to
follow. Nothing on a CPU node reaches the kernels this PR changes.

anko9801:
**Bitwise check: the CPU speedup changed no numbers.**

I had been *arguing* the three CPU changes are exact rather than showing it, so
`bench/itp_state_fingerprint.jl` runs a 40-step Eu ITP (16³, DDI padded, all
three changed paths live) and hashes the raw bits of ψ. Both worktrees, one
script file, one node, one job (`scripts/tsubame/submit_itp_fingerprint.sh`):

```
###### baseline  950f9b3b
psi_hash   = 0x49868494f37f5a85
energy     = 11.707083885123401
###### optimised 2d99583f
psi_hash   = 0x49868494f37f5a85
energy     = 11.707083885123401
```

Identical to the bit. `tol = 0` so the run cannot converge early and all 40
steps execute; `converged = false, last_step = 40` on both.

GPU gates are still queued on gpu_1 (~25 min so far). I did tighten one
assertion in `test_gpu_padded_corner_parity.jl` before resubmitting: comparing
the fused kernel to `_spin_density_corner_bcast!` with `==` was asserting a
property of NVPTX's fma contraction, not of this code, so that half is now
machine precision. `==` stays for padded-vs-contiguous, which is the same
kernel with a different index and is the assertion that actually constrains the
map.

anko9801:
**A fourth casualty of the same default flip, larger than the three this PR fixes — and NOT fixed here.**

Found while auditing the call sites of the symbols I changed.
`_spin_chain_reason` (`src/hamiltonian/integrator/spin_chain.jl:64`) disqualifies
the fused `diag · SM · DDI · SM · diag` half-step outright:

```julia
ws.ddi_padded === nothing || return "zero-padded DDI uses a different convolution"
```

Dealias defaults OFF, and the other exclusions (c₂, tensor, Raman, light shift,
transverse Zeeman, tabulated LHY, magnetic gradient) do not apply to a plain Eu
`c₀/c₁` + DDI run — so that shape **was** eligible before 9c117c05 and is not
now. Every production RTP step since this morning takes the operator-by-operator
chain instead of the one-pass fused kernel that file exists to provide.

I have not fixed it, deliberately: it is the RTP path, not ITP, and the fix is
to teach the fused kernel the padded convolution rather than to relax the
eligibility test — the guard is correct as written, because `gpu_spin_chain.jl`
reads the *unpadded* `bufs.Phi_*`, which the padded convolution never fills.
Relaxing it without porting the kernel would be a silent correctness bug, not a
speedup. Worth its own issue.

(This is also why the `_gpu_phi_vec` crop left in `gpu_spin_chain.jl` is not a
loose end from this PR: that path only ever runs unpadded, so it always takes
the zero-copy branch.)

anko9801:
**GPU gates: green.** H100 (r6n5), commit 545cf1df.

| gate | result |
|---|---|
| `gpu/test_gpu_padded_corner_parity.jl` (new) | **32 / 32** |
| `gpu/test_gpu_spin_rotation_taylor_parity.jl` | **48 / 48** (+ 8 + 8 + 8 + 2 sub-suites) |
| `oracles/test_gpu_cpu_per_term_parity.jl` | **156 / 156** |

That closes the last open item. Full verification for this PR:

* GPU kernels — the three gates above, on an H100. Nothing on a CPU node reaches
  them, so this is the only thing that covers the changed CUDA code.
* CPU numerics — ψ bit-identical to baseline (`psi_hash = 0x49868494f37f5a85`,
  energy to 17 digits), so the CPU speedup provably moved nothing.
* Whole suite — `ci` tier, 258 files, all passed.
* Targeted CPU gates — 8 files covering DDI / padded DDI / spin-mixing /
  ITP-DDI-Strang / translation covariance / spin-density consistency.

Speedup, unchanged from the top comment: **2.02× on H100 at 32³**
(0.992 → 0.490 ms/step), −18.4 % at 64³, −8.3 % / −11.4 % on cpu_4 at 32³ / 48³.

Ready for review.

PR #185, opened 2026-07-29
fix(lbfgs): report a real dE, and leave the iterate the line search measured

anko9801:
Two bookkeeping defects in `find_ground_state_lbfgs`, found while reading the solver's hot path for a separate performance change. Split out so they can land on their own.

## `result.dE` was identically zero

`E_prev` was assigned `E_trial` — the energy **at** the accepted point. That is the same iterate `E_new` is measured at, and bit-identical to it, so `dE = abs(E - E_prev)` differenced a value against itself. From step 2 onward, every run reported `dE == 0.0`, including the value returned in the result NamedTuple and the one printed by `verbose`.

`E_prev` is now the energy before the step, so `dE` is the step's energy decrease.

## The line search could return an energy for a state it did not leave

The expansion phase guards its re-placement with `best_α == α_init`. That is true precisely in the common case where the **first** trial doubling is rejected — and in that case `ws.state.psi` was left holding the rejected step while the returned `best_E` described `α_init`. The function's own comment says it leaves the workspace at best.

Latent at HEAD: the driver recomputes the retraction and `energy_gradient!` overwrites `ws.state.psi`, so nothing observes it today. It is fixed here because it is a mismatch between a returned value and reachable state, and any caller that starts trusting the returned state would inherit it silently.

The guard is now on the last evaluated α.

## Gate

`test/solvers/test_lbfgs_line_search_and_de.jl`:

- the `dE` assertion is an identity, not a sign check: the trajectory is deterministic, so a 6-step run's last step is exactly the 5-step run's endpoint, and the 6-step run's `dE` must equal the gap between the two runs' energies. `dE > 0` alone would flake on a step whose line search fails.
- the line-search assertion sweeps the direction scale so the accept-then-expand branch, the backtracking branch, and the "first doubling rejected" corner are all exercised, and asserts `total_energy(ws) == E` exactly for every accepted return.

It is registered in `CI_EXTRA`, not `FULL_EXTRA`. Every other L-BFGS test lives in `FULL_EXTRA`, so a green `ci` run currently executes none of them — a `tier=ci` run on this repo passed 257 files without touching the solver.

## Verification

`tier=ci` on TSUBAME (job 8297433) — will report below before merging.

### DIFF
--- a/src/solvers/lbfgs/driver.jl
+++ b/src/solvers/lbfgs/driver.jl
@@ -303,10 +303,15 @@ function find_ground_state_lbfgs(;
 
         # Carry the new gradient/energy to the next iteration (swap buffers so the
         # old `grad` is reused as scratch for the next `grad_new`).
+        # `E_prev` is the energy BEFORE this step, so `dE` is the step's energy
+        # decrease. It used to be assigned `E_trial` — the energy AT the accepted
+        # point, i.e. the same iterate `E_new` is measured at, and bit-identical
+        # to it — so the reported `dE` was exactly 0.0 from step 2 onward, for
+        # every run, including the value returned in `result.dE`.
         grad, grad_new = grad_new, grad
         grad_norm = grad_norm_new
+        E_prev = E
         E = E_new
-        E_prev = E_trial
         last_step = step
     end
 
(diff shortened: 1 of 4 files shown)


anko9801:
**Verification (TSUBAME, job 8297433):** `tier=ci` at `d7910a1f` — **259 files, all passed**, including the new gate `solvers/test_lbfgs_line_search_and_de.jl` (15/15, 39 s).

That tier matters here: it is the one that previously ran 257 files without executing a single L-BFGS test, because every other L-BFGS test lives in `FULL_EXTRA`. The new gate is registered in `CI_EXTRA` so it actually runs.

`tier=full` on this repo is red on `main` independently of this PR (7 files, incl. `test_lbfgs_accuracy_floor.jl:118`); this branch does not add to that set.

Issue #184, opened 2026-07-29
The zero-padding default flip disabled the fused spin-chain half-step for every production RTP run


anko9801:
`_spin_chain_reason` (`src/hamiltonian/integrator/spin_chain.jl:64`) refuses the
fused `diag · SM · DDI · SM · diag` half-step whenever a padded DDI context
exists:

```julia
ws.ddi_padded === nothing || return "zero-padded DDI uses a different convolution"
```

`DDI_PADDED_DEFAULT` flipped to `true` in 9c117c05 (2026-07-29), so every
YAML-driven run now builds `ddi_padded` and the fusion is off unconditionally.

**The guard is correct as written — do not relax it.** `gpu_spin_chain.jl` reads
the *unpadded* `bufs.Phi_{x,y,z}`, which the padded convolution never fills, so
allowing the fused path through with padding on would be a silent correctness
bug rather than a speedup. The fix is to teach the fused kernel the padded
convolution (it already needs only the `[1:n_pts...]` corner of Φ, which
`_voxel_index` in `src/foundation/voxel_index.jl` now expresses).

**Scope of the loss.** Dealias defaults OFF and none of the other exclusions
(c₂, tensor, Raman, light shift, transverse Zeeman, tabulated LHY, magnetic
gradient) apply to a plain ¹⁵¹Eu `c₀/c₁` + DDI run — the exact production shape.
That shape was eligible before the flip and is not now. `spin_chain.jl` describes
the fusion as bit-identical to the operator-by-operator chain, i.e. this is pure
throughput, and the file exists because the win is worth having.

**Not measured.** I found this by auditing call sites while working on the ITP
hot path (#183) and have not benchmarked what the fusion was buying at the
current defaults; the number in this issue should be measured before deciding
priority. Related same-flip findings, with numbers, are in #183 — the ITP-side
casualties were worth 2.02× on an H100 at 32³.

Suggested first step: run `bench/profile_1step.jl` (RTP) with
`ddi_padding=false` vs `true` at a production shape to size the gap before
committing to the port.

anko9801:
Fixed on main by #192 (`b3f07bd8 fix(rtp): a padded DDI declined the fused V half-step, so no production run took it`), and #193 then cut traffic in the fused half-step on top of it. Not my work — another session picked it up.

Closing as resolved. For the record, the fix took the route this issue argued for
— teaching the fused kernel the padded convolution rather than relaxing the
eligibility guard, which would have been a silent correctness bug since
[...]

PR #191, opened 2026-07-29
perf(itp): 2.2× on CPU — Taylor-Horner spin rotation, and drop the per-step host sync

anko9801:
Two independent ITP wins, measured on TSUBAME with both arms in ONE job on one
node/card. Baseline is `main` (634e0234), i.e. after #183.

| config | before | after | Δ |
|---|---|---|---|
| cpu_4 32³ | 60.43 ms | **27.56 ms** | **2.19×** |
| cpu_4 48³ | 244.2 ms | **108.4 ms** | **2.25×** |
| H100 32³ | 0.506 ms | **0.472 ms** | −34 µs (−6.7 %) |
| H100 64³ | 2.645 ms | **2.624 ms** | −21 µs (−0.8 %) |

Against the pre-#183 baseline the CPU step is now **2.39×** at 32³ and **2.48×**
at 48³.

## 1. CPU spin rotation by Taylor–Horner (the large one)

The spin rotation was ~67 % of a CPU ITP step — `spin_mixing` 4 × 6.97 ms and
`ddi_rotation` 2 × 6.33 ms of a 60.4 ms step at 32³. The Euler 5-stage moves ψ
through cache ~18 times per rotation (5 phase passes + 4 gemms) and needs an
`atan`/`acos`/`sqrt` pre-pass to turn the field into Euler angles.
Taylor–Horner moves it twice and needs no trigonometry at all: the spinor sits
in stack buffers for the whole rotation and the tridiagonal matvec is written
out. Per substep:

| | before | after |
|---|---|---|
| `spin_mixing` (32³) | 6.97 ms | **1.48 ms** (4.7×) |
| `ddi_rotation` (32³) | 6.33 ms | **1.34 ms** (4.7×) |
| `spin_mixing` (48³) | 26.8 ms | **4.60 ms** (5.8×) |
| `ddi_rotation` (48³) | 24.6 ms | **4.23 ms** (5.8×) |

`ddi_convolve` and `diagonal` are unchanged, as they should be.

This is the realization the GPU has had since its per-voxel-degree rewrite
(3.6× against Euler at D=13 there); the CPU had no counterpart. It is a
**different realization, not a port** — the GPU form is warp-cooperative, one
spin component per lane with neighbours by `shfl`, which a CPU core has nothing
to match.

Three things stop being stated twice:

* **the accuracy contract** — `SPIN_TAYLOR_{ENABLED,TOL,RSAFE,RK_MAX}` move to
  `src/foundation/spinor_utils/spin_rotation_taylor.jl` and the CUDA ext reads
  them, so the two devices cannot drift to different tolerances;
* **the generator** — `spin_tridiag_bands` reads the bands off `sm.Fx/Fy/Fz`
  and both devices call it;
* **the spin-mixing field** — the rotation axis now comes from
  `_compute_spin_density!` rather than a second hand-written copy of that
  reduction fused into the angle pre-pass.

### This one is NOT bit-identical

It is a truncated series with a per-voxel degree, targeting a backward error of
`SPIN_TAYLOR_TOL = 1e-13` per rotation. That is a deliberate accuracy decision
and it changes every CPU ITP/RTP result at that level, so it is gated
accordingly:

`test/hamiltonian/test_cpu_spin_rotation_taylor_parity.jl` compares against the
**exact Euler form running on the same input** — deliberately not against the
GPU, because CPU and GPU would otherwise agree to the same wrong tolerance. It
sweeps F ∈ {1,2,6}, real/imaginary time, and R from 0.01 (well inside the
0.01–0.2 production range) to 30 (far past `SPIN_TAYLOR_RSAFE`, where every
voxel halves its angle). It also pins real-time unitarity over 20 rotations and
asserts the per-voxel stack buffers do not reach the heap — an escaping
`MVector` at D=13 would hand back exactly the traffic this kernel saves.

`SPIN_TAYLOR_ENABLED[] = false` restores the exact Euler path.

## 2. The ITP normalisation no longer syncs to the host

`sum(abs2, psi)` returns a **host** scalar, so it ends in a device-to-host copy
— a full stream drain. ITP normalises on every step (`normalize_every = 1`
whenever the magnetisation is unconstrained). Reducing with `dims` leaves the
result in a `(1,1,…,1)` device array that broadcasts against ψ.

Worth stating plainly: I predicted this from the isolated kernel timing (61 µs,
"12 % of a 32³ step") and it delivered **34 µs, 6.7 %**. The isolated
measurement overstated it because the benchmark syncs around each part anyway;
the honest number is the whole-step one in the table.

## Verification

* `ci` tier and GPU gates: see the comment below.
* Targeted CPU gates, all green on TSUBAME:
  `test_cpu_spin_rotation_taylor_parity`, `test_spin_mixing`, `test_ddi`,
  `test_ddi_padded`, `test_itp_ddi_strang_save_every`,
  `test_spin_density_consistency`, `test_ddi_translation_covariance`,
  `analysis/test_spin_rotation`.

## Still not done

`#184` — the padding default-flip also disabled the fused RTP spin-chain
half-step. Untouched here; that is the RTP path and needs the fused kernel
taught the padded convolution.

### DIFF
--- a/bench/profile_1step_gpu.jl
+++ b/bench/profile_1step_gpu.jl
@@ -104,13 +104,13 @@ t_diag = gtime(() -> SpinorBEC._dispatch_diagonal_step!(ws, Val(3), zd, 0.00125,
 let Ext = Base.get_extension(SpinorBEC, :SpinorBECCUDAExt)
     pm = sqrt(maximum(bufs.Phi_x.^2 .+ bufs.Phi_y.^2 .+ bufs.Phi_z.^2))
     R = 0.0025 * pm * Float64(sm.system.F)
-    use0 = Ext._SPIN_TAYLOR_ENABLED[]
+    use0 = SpinorBEC.SPIN_TAYLOR_ENABLED[]
     @printf("\nDDI rotation: peak R = dt·max|Φ|·F = %.4f (the kernel sizes each voxel from its own R)\n", R)
-    Ext._SPIN_TAYLOR_ENABLED[] = true
+    SpinorBEC.SPIN_TAYLOR_ENABLED[] = true
     t_rot_tay = gtime(() -> SpinorBEC._apply_ddi_rotation!(psi, bufs.Phi_x, bufs.Phi_y, bufs.Phi_z, sm, 0.0025, 3))
-    Ext._SPIN_TAYLOR_ENABLED[] = false
+    SpinorBEC.SPIN_TAYLOR_ENABLED[] = false
     t_rot_eul = gtime(() -> SpinorBEC._apply_ddi_rotation!(psi, bufs.Phi_x, bufs.Phi_y, bufs.Phi_z, sm, 0.0025, 3))
-    Ext._SPIN_TAYLOR_ENABLED[] = use0          # restore the real default
+    SpinorBEC.SPIN_TAYLOR_ENABLED[] = use0          # restore the real default
     @printf("  ddi_rotation  Taylor %8.1f μs   Euler %8.1f μs   (Euler/Taylor=%.2fx)\n",
         t_rot_tay, t_rot_eul, t_rot_eul / t_rot_tay)
 end
--- a/bench/rtp_gpu_ab.jl
+++ b/bench/rtp_gpu_ab.jl
@@ -42,7 +42,7 @@ function build(n; c1_ratio=0.05, dt=1e-4)
 end
 
 function arm!(ws, psi0, mode::Symbol, k)
-    Ext._SPIN_TAYLOR_ENABLED[] = mode !== :euler
+    SpinorBEC.SPIN_TAYLOR_ENABLED[] = mode !== :euler
     step! = mode === :combined ? SpinorBEC.split_step_combined! : SpinorBEC.split_step!
     copyto!(ws.state.psi, psi0)
     ws.state.t = 0.0
@@ -97,7 +97,7 @@ for n in SIZES
 end
 
 # Kernel breakdown of both surviving paths at the largest size.
-Ext._SPIN_TAYLOR_ENABLED[] = true
+SpinorBEC.SPIN_TAYLOR_ENABLED[] = true
 let n = SIZES[end]
     for (label, step!) in (("sequential", SpinorBEC.split_step!),
         ("combined", SpinorBEC.split_step_combined!))
--- a/bench/taylor_tolerance_sweep.jl
+++ b/bench/taylor_tolerance_sweep.jl
@@ -94,7 +94,15 @@ println("="^78)
 
 # Reference 1: the EXACT rotation. Any Taylor arm's deviation from this is
 # truncation error and nothing else.
-println("\n[1/3] exact Euler reference…")
+# Warm up BOTH realizations before anything is timed. The first version of this
+# script timed the exact-Euler reference first and cold, which made its wall
+# time absorb the whole JIT cascade and reported a "4×" whole-run speedup that
+# was an artefact. Kernel-level A/B (bench/bench_itp_step.jl) said 2.19×.
+println("\n[0/3] warm-up (both realizations)…")
+run_itp(; dt=DT, n_steps=2, taylor=false)
+run_itp(; dt=DT, n_steps=2, taylor=true)
+
+println("[1/3] exact Euler reference…")
 ref = run_itp(; dt=DT, n_steps=N_STEPS, taylor=false)
 @printf("  E_exact   = %.15g   (%.1f s)\n", ref.E, ref.wall)
 
--- a/bench/taylor_tolerance_sweep.jl
+++ b/bench/taylor_tolerance_sweep.jl
@@ -1,0 +1,131 @@
+# What accuracy does the Taylor spin rotation actually need, and what does it
+# cost?
+#
+# `SPIN_TAYLOR_TOL` was inherited from the CUDA kernel, where it was justified by
+# a measurement on that device (1e-9 → 1e-13 cost 1.088× at 64³ and 1.024× at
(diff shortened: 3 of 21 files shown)


anko9801:
## The tolerance measurement, and what it changed

@ the accuracy question: `SPIN_TAYLOR_TOL = 1e-13` was **inherited from the
CUDA kernel without being measured on the CPU**. That was a hole in this PR —
the GPU's justification (memory-bound, so the degree is nearly free) is a
statement about that device.

`bench/taylor_tolerance_sweep.jl` measures it against the thing that actually
binds the answer: the splitting error the caller already accepted by choosing
`dt`. Eu F=6, 32³, dt = 0.002, |E(dt) − E(dt/2)| = **7.6e-3**:

| tol | \|ΔE\| vs exact | fraction of splitting error | mean degree |
|---|---|---|---|
| 1e-5 | 2.5e-7 | **3.3e-5** | 2.00 |
| 1e-9 | 2.9e-12 | 3.9e-10 | 2.26 |
| 1e-13 | 2.4e-13 | 3.1e-11 | 2.65 |
| 1e-15 | 2.4e-13 | 3.2e-11 | 2.82 |

The answer saturates at 1e-11; the degree moves 2.00 → 2.82 across ten decades;
and **even at the loosest tolerance tested the rotation contributes 4.5 orders
of magnitude less error than `dt` does.** So the tolerance is not
free-but-chosen — over the whole tested range it is irrelevant. That is the
argument for not exposing it: there is no setting of it a caller could benefit
from knowing about.

### The positive control found my gate was vacuous

I first gated this with a control that loosened `tol` and expected the
criterion to break. **It cannot break that way.** The degree is floored at 2, so
at production R ≈ 0.01–0.2 the schedule returns degree 2 for every tolerance
from 1e-15 up to 1 — visible in the table as the same mean degree 2.00 at 1e-5
and 1e-7 — and that floor alone already passes. The control would have failed
while the code was correct, because it asserted something unreachable.

The criterion is held by the **degree floor**, not by the tolerance. The control
now attacks the floor: an order-1 truncation of the same rotation, built from
the same bands so the comparison isolates the order, must be materially worse.
The floor is pinned separately across g and tol.
`test_taylor_tolerance_criterion.jl`: **15/15**.

### Two corrections to my own numbers

* The sweep's `vs Euler` column originally read 0.23–0.26, i.e. a "4× whole-run
  speedup". That was an artefact: the exact-Euler reference ran first and cold,
  absorbing the JIT cascade. Warmed, it reads **0.44–0.47**, which reconciles
  with the kernel-level **2.19×**. Both realizations are warmed before timing now.
* For the normalisation change I predicted 12 % of a 32³ step from the isolated
  kernel timing; the whole-step A/B gives **6.7 %**. The table in the PR body is
  the whole-step number.

### How far the margin reaches

The ratio is not scale-free: splitting error falls as dt² while truncation
accumulates over ~1/dt rotations, so truncation/splitting grows roughly as
**dt⁻³** — about three decades of margin per factor of ten in dt. Production dt
is 1e-3–5e-3 so 3e-11 is comfortable, but the number belongs to the dt it was
established at. Recorded in the gate's header rather than left implicit.

`ci` is re-running on the final branch state; previous run was 259/259.

anko9801:
## The method became a mechanism, and it caught me twice

@ "それを作ろうか" — `src/workflow/validation/error_budget.jl`.

`ErrorBudget` records three errors on ONE observable and ONE config: the
**baseline** (what the caller's own choice already costs — e.g.
`|obs(dt) − obs(dt/2)|` with the approximation off), the **approximation**'s
contribution, and a deliberately degraded **control**.
`check(NegligibleErrorSpec(frac), budget)` gives a verdict and **refuses to give
one when it could not mean anything**:

* `:indeterminate` — the baseline is degenerate, **or the control does not
  breach**. In the second case a `:pass` would say "this comparison cannot
  fail", not "the approximation is good".
* `:fail` / `:pass` — otherwise.

`CheckResult` already carried `:indeterminate` for "precondition unmet". This is
that mechanism applied to the precondition nobody states out loud.

### It rejected two of my own gates before one passed

1. **Control = loosen `SPIN_TAYLOR_TOL`.** Cannot breach: the Horner degree is
   floored at 2, and at production `R = |scale|·|v|·F ≈ 1e-5` the schedule
   returns 2 for every tolerance over ten decades.
2. **Control = lower the degree floor to 1.** Also cannot breach: an order-1
   rotation is still ~1e-8 relative, four orders inside the splitting error. So
   the criterion is **not held by the floor either**, which is what I had
   claimed one commit earlier and have retracted.

What actually holds it is that `R` is tiny — the rotation is not the binding
error at any order ≥ 1. The only remaining control is to **remove the operator**
(`SPIN_TAYLOR_DEGREE_CAP[] = 0` skips the Horner loop). That is also the honest
general form: show the observable moves when the operator is absent, or
"negligible" was never a claim about anything.

Final verdict, `test_taylor_tolerance_criterion.jl` **13/13**:

```
approximation 1.75e-14  <  1e-3 × baseline = 9.67e-8   (ratio 1.8e-10)
control (rotation removed) breaches at 4.61e-2
```

**Both wrong controls would have shipped as green gates asserting nothing.** I
would have reported "the criterion holds" with confidence, twice, and been
wrong. That is the argument for the mechanism, and it is not a hypothetical one.

### What it does not do

Documented in the header, because the tool cannot check any of it: it needs an
exact reference; a baseline that genuinely dominates **and** was genuinely
chosen by the caller; and the observable you actually care about — negligible in
energy is not negligible in a winding number. The ratio is also not scale-free
(≈ dt⁻³ here), so a budget belongs to the config it was measured at.

GPU gates and `ci` are re-running on the final commit (the CUDA schedule's
argument changed with the cap). Previous states: `ci` 260/260, GPU 32/32 +
48/48 + 156/156.

PR #194, opened 2026-07-29
perf(lbfgs): 1.6-2.4x per iteration, by not recomputing what the iteration already has

anko9801:
Makes `find_ground_state_lbfgs` **1.62–2.37× faster per iteration on CPU**, measured. Depends on nothing outside the solver; the two behavioural bugs found along the way landed separately in #185, which this branch merges.

## Measured

TSUBAME, **one UGE job on one node**, both revisions back to back (an A/B split across two jobs would put a node-to-node and queue-epoch confound into the only number this is judged by). `gpu_h`, `JULIA_NUM_THREADS=8`, Eu-151 F=6, D=13, box 12, `m_lbfgs=20`. Wall per L-BFGS iteration as the *slope* of wall against `n_steps`, so setup/JIT/finalisation drop out.

| cell | CPU before | CPU after | |
|---|---|---|---|
| 16³ contact | 15.40 ms | **6.50 ms** | 2.37× |
| 16³ +DDI | 16.74 ms | **9.09 ms** | 1.84× |
| 24³ contact | 49.11 ms | **27.47 ms** | 1.79× |
| 24³ +DDI | 52.43 ms | **32.30 ms** | 1.62× |

**The GPU column is deliberately not claimed.** Four independent jobs put the same baseline cells 10–19 % apart — comparable to the differences — and the sign of one cell is not stable across jobs. Numbers are in the CSV; the figure shows both backends.

Figure + data: `docs/figs/lbfgs_iteration_cost_ab.{png,csv}`. Narrative and cost model: `docs/design/lbfgs_iteration_cost.md`.

## What was removed

Everything here is work whose answer the iteration already had, or work it was doing per call that it could do once.

- **A second total-energy evaluation per iteration.** The line search evaluates `total_energy` at the step it accepts; the gradient pass then recomputed the same energy at the same ψ — on the CPU a second full traversal of the term registry (kinetic FFTs, DDI convolution, spin density). The line search now returns its accepted iterate, the driver adopts it instead of redoing step + retraction, and the gradient goes through a new `gradient_only!` with the known energy passed in.
- `_finalize_lbfgs_atomic!` called `total_energy` and then `energy_gradient!`, which returns that same total at that same ψ.
- **Per-iteration allocations.** `push!(s_hist, copy(s_k))` allocated two ψ-sized arrays every iteration and dropped two more; at capacity the evicted buffers are now reused. `s_k`/`y_k` come from the scratch registry.
- **Three per-component FFT loops** — the Sobolev preconditioner, its forward metric, and the combined `P_V^½ P_K P_V^½` preconditioner had each independently grown the same `copy slice → fft → scale → ifft → copy back`. They now share one `batched_kspace_filter!`: one batched plan pair, in place, zero copies. GPU Sobolev 0.46 → 0.035 ms.
- **A host→device copy of k² per call.** `Grid.k_squared` is a host `Array` even for a GPU workspace, and the kinetic operator face ran `_to_device` on every call — a fresh `CuArray` plus a full transfer per gradient and per energy evaluation.
- **GPU per-call allocations**: `_gpu_energy_and_optional_grad` took its per-term buffer from `similar(psi)` on every call (and it is *also* the GPU `energy_decomposition`, so once per line-search trial); `build_gradient_context` took the device density through the allocating `total_density`; MagneticGradient was evaluated unconditionally.
- **The two-loop recursion**, which the measurement showed was 55–71 % of the CPU iteration: 9.9 ms at 16³, 31.5 ms at 24³, at ~18 GB/s — one core's streaming bandwidth. It is pure history traffic (`2m` ψ-sized arrays read twice each) and an F=1 contact problem pays it exactly as much as Eu F=6 does. Now 2.2 / 6.8 ms.

## Accuracy

Two changes touch arithmetic, and are gated accordingly.

`_axpy_threaded!` splits the two-loop's axpys across threads. Elementwise, so the result cannot depend on the split — gated **bit-identical** to the broadcast it replaces, at sizes either side of its threshold, and identical by construction at `nthreads() == 1`.

`_realdot_blocked` replaces `real(dot(a,b))` with a fixed-64-block reduction. This one is **not** bit-identical to BLAS `zdotc`, and that is stated rather than buried: the solver sits close enough to the `√eps` energy-gated floor that any change in summation order moves the endpoint. The gate measures it against a **BigFloat reference on a deliberately ill-conditioned input** (a planted cancelling pair, condition ~1e10) and requires it to be no worse than a sequential sum — blocking is one level of pairwise summation, so the error bound improves from `O(n·eps)` to `O((n/64 + 64)·eps)`. It is also independent of the thread count, which a `@threads`-over-`nthreads()` reduction would not be.

Worth recording: the **first** version of that dot was hand-unrolled scalar, chosen to be machine-independent, and measured **8.20 ms against `zdotc`'s 7.72 ms** *with its blocks threaded* — the scalar loads alone cancelled four cores. "Blocked and threaded" sounds like it must be faster and was not. `@simd` is what makes each block memory-bound.

Everything else is gated as bit-identical against an independent re-statement of the loop it replaced (`test/solvers/test_lbfgs_fast_path_equivalence.jl`).

## Verification

- `tier=ci` at `c75671f5`: **262 files, all passed** (TSUBAME job 8297628).
- `tier=full` running; will report below. On the pre-merge tip it failed 7 files, **all 7 of which also fail on `main`** — `main` fails one more, so the branch's failure set was a strict subset.

## Not in this PR

- `_grad_ddi!` has no padded branch while `_ddi_energy` does, so since `ddi_padding` defaulted ON a DDI ground state descends an *unpadded* gradient toward a *padded* energy. Correctness, not performance; changes every padded-DDI ground state, so it wants its own PR. `test_ddi_padded.jl` never calls `energy_gradient!` and the FD oracle builds its workspaces without padding, which is why nothing caught it.
- A CPU fused energy+gradient. `α = 1` is accepted on essentially every iteration, so fusing the line search's first trial with the gradient would replace an energy pass plus a gradient pass with one pass — the largest single item left. It needs per-term fusion (share the forward FFT between `_kinetic_energy` and `_grad_kinetic!`, share the DDI potential) rather than switching the energy faces to `factor·Re⟨ψ,Hψ⟩`, so that no energy value changes.


anko9801:
**Verification (TSUBAME):**

- `tier=ci` at `c75671f5` — **262 files, all passed** (job 8297628).
- `tier=full` at `c75671f5` (job 8297629) — 7 workers with failures:

```
gpu/test_mixed_precision.jl:64
gpu/test_mixed_precision_phase3.jl:51
hamiltonian/test_combined_spin_step.jl:146,148
oracles/test_spin_chain_fusion_parity.jl:69
solvers/test_yoshida_ddi_order.jl:74,75,83,87
workflow/test_phase_diff_eval.jl:75
workflow/test_pipeline.jl:382
```

All except `test_yoshida_ddi_order.jl` also fail on a clean baseline worktree (measured at `550fe24e` = `main` before this branch: same set, plus `test_lbfgs_accuracy_floor.jl:118`, which passed this time — see below).

`test_yoshida_ddi_order.jl` is new relative to that baseline, but the baseline predates several `main` commits this branch now merges (#183 `perf/itp-step` among them, which touches the ITP/propagator path this test measures the convergence order of). A `tier=full` run on `origin/main` HEAD is in flight (job 8297857) to settle whether it is this branch or `main`; result below before this is merged.

Two notes on `tier=full` on this repo generally:

- `solvers/test_lbfgs_accuracy_floor.jl:118` ("Newton-CG polish tightens the gradient floor") **fails on clean `main`**, where the polish returns a **bit-identical** ψ — every trust-region step rejected, so the pass is a no-op — and the L-BFGS floor it is compared against is not reproducible: 5.7e-10, 1.4e-8 and 2.8e-8 measured for the same commit and problem, while the energy is bit-exact at `E − 0.5 = −2.2e-16`. It passed on this run, which is the same statement.
- `workflow/test_phase_diff_eval.jl:75` appeared on the baseline, was absent from the pre-merge run of this branch, and is back here. Also not stable.

anko9801:
**Settled: `test_yoshida_ddi_order.jl` fails on clean `main` too.**

`tier=full` on a clean `origin/main` worktree at `bdc8aec8` (job 8297857) — **8/12 workers with failures**:

```
gpu/test_mixed_precision.jl:64
gpu/test_mixed_precision_phase3.jl:51
hamiltonian/test_combined_spin_step.jl:146,148
oracles/test_spin_chain_fusion_parity.jl:69
solvers/test_lbfgs_accuracy_floor.jl:118          <- not present on this branch's run
solvers/test_yoshida_ddi_order.jl:74,75,83,87
workflow/test_phase_diff_eval.jl:75
workflow/test_pipeline.jl:382
```

This branch's set at `c75671f5` is **exactly that, minus `test_lbfgs_accuracy_floor.jl:118`** — a strict subset. Nothing here adds a failure.

Summary of verification:

| | `tier=ci` | `tier=full` |
|---|---|---|
| `origin/main` @ `bdc8aec8` | — | 8 workers fail |
| this branch @ `c75671f5` | **262 files, all passed** | 7 workers fail, ⊂ main's set |

anko9801:
## ⚠️ Numbers withheld — converting to draft

**Do not merge on the table in the PR description.** All four A/B rounds ran on `gpu_h`, which is *half a node*: the other tenants share the socket's memory controllers. The two-loop recursion is memory-bandwidth bound, so that allocation cannot measure it, and the baseline arm always ran before the optimised arm — a monotonic drop in someone else's load produces the same picture as a speedup.

The tell came from the thread scan (`node_q`, 24³, `t_dir` = the two-loop):

| threads | 1 | 2 | 4 | 8 | 16 | 32 |
|---|---|---|---|---|---|---|
| two-loop, ms | 8.34 | 15.67 | 14.23 | 11.37 | 8.26 | 12.28 |

Non-monotonic, and **no faster at 32 threads than at 1**. On `gpu_h` the same component read 31.5 ms for the baseline and 6.8 ms for this branch, and 4.6× of that was attributed to threading the axpys. At 1 thread here — every threaded branch off — it is 8.3 ms. That attribution does not survive.

Re-measuring on `node_f` (full node, exclusive, so nothing else touches the memory system), job 8300905: the A/B for both arms **and a thread scan of both arms**, so the attribution is measured rather than assumed.

## What this does and does not put in doubt

Most of this branch removes work outright, and that argument is from the code, not from a timer:

- the line search's accepted energy was being recomputed by the gradient pass — one full registry traversal per iteration, gone
- `_finalize_lbfgs_atomic!` evaluated the same total twice
- four ψ-sized allocations per iteration → zero at steady state
- three preconditioners: `2·D` slice copies → zero, `2·D` plan calls → 2
- the kinetic face's host→device copy of k² per call → once per grid
- the GPU per-term buffer's `similar(psi)` per call → cached
- γ's dot product → read from `rho_hist`

In doubt is exactly the threading pair — `_axpy_threaded!` and `_realdot_blocked`. If the exclusive-node scan says they buy nothing, **they come out**: the first adds a code path for no gain, and the second also gives up bit-identity with BLAS `zdotc`, which is not a price worth paying for nothing.

PR #195, opened 2026-07-29
fix(io): result.jld2 and point_NNN.jld2 each carried the whole snapshot payload

anko9801:
Found while investigating why the TSUBAME personal work area sat at 161 GB against a 100 GiB quota.

## The duplication

`run_pipeline` auto-saves the dashboard-canonical `result.jld2` at its end (`runner.jl:217`), full streamed snapshot payload included, and symlinks `point_001.jld2` at it — with the comment *"so we don't duplicate the multi-GB snapshot data"*. Then `_run_yaml_single` writes the real `point_NNN.jld2` **over that symlink**. The intent is defeated by ordering and both files end up carrying every frame.

Measured on a live Eu movie run:

```
one 753-frame dynamics:  point_001.jld2  10.25 GB
                         result.jld2      9.71 GB
23 such pairs:                           83.7 GB of pure duplication
```

Still growing — a run that finished 30 minutes before I looked had just written 11.12 + 10.43 GB.

## Verified redundant before touching anything

- snapshot key sets identical (753 vs 753)
- 12 sampled frames **bit-identical**
- `times` / `Fz` / `norms` bit-identical
- `result.jld2`'s key set is a **strict subset** of the point file's
- every directory holding a `result.jld2` also holds a `point_001.jld2` (0 exceptions across 23)

## The fix

Collapse it the **other way round** from the existing attempt: keep the point file — the richer one (analyze results, per-term energy, interactions, units, env) — and make `result.jld2` a **relative symlink** to it. Everything that opens `result.jld2` keeps working and gets a superset.

**Not unconditional.** `save_rotating_basis_result!` writes `Lz`, `per_m_history` and `integrator_meta` for rotating-basis runs, which the point file does *not* carry — collapsing blindly would lose data. `_collapse_duplicate_result_jld2` compares key sets (top level and one level into `dynamics`) and leaves both files alone unless the point file provably covers `result.jld2`. Keys only, no frame is read, so it costs milliseconds.

## Gate

`test/workflow/test_result_jld2_dedupe.jl` — 18 assertions over four cases:

- collapses on a superset, and the link is **relative** so the directory can move
- **refuses** on the rotating-basis shape where `Lz` / `per_m_history` are result-only
- refuses on a top-level-only difference
- inert when already collapsed, or when `result.jld2` is absent

## Not in this PR

Existing duplicates are untouched. They sit next to jobs that are currently writing, so the backlog sweep needs its own interference-safe pass — completed runs only, skip directories where `point_001.jld2` is itself a symlink, verify per directory before deleting.

Also adds a `_COST` entry the runner asked for: `test_interactions_constraint.jl` had none, so it fell back to the 3.0 default while measuring 18.2 s.

`tier=fast`: 169 files, all passed.

### DIFF
--- a/src/workflow/experiments/pipeline/run_registry.jl
+++ b/src/workflow/experiments/pipeline/run_registry.jl
@@ -694,6 +694,60 @@ function _snapshot_compression_kwargs(result)
     return (;)
 end
 
+# `run_pipeline` auto-saves the dashboard-canonical `result.jld2` at its end
+# (runner.jl), including the full streamed snapshot payload, and symlinks
+# `point_001.jld2` at it — with the comment "so we don't duplicate the multi-GB
+# snapshot data". `_run_yaml_single` then writes the real `point_NNN.jld2` over
+# that symlink, so the intent is defeated by ordering and BOTH files end up
+# carrying every frame. Measured on one Eu movie run: point_001 10.25 GB +
+# result 9.71 GB for one 753-frame dynamics, and 23 such pairs had accumulated
+# to 83.7 GB of pure duplication.
+#
+# Collapse it the other way round: keep the point file (it is the richer one —
+# analyze results, per-term energy, interactions, units, env) and make
+# `result.jld2` a symlink to it.
+#
+# ONLY when the point file provably contains everything `result.jld2` does.
+# That is not automatic: `save_rotating_basis_result!` writes `Lz`,
+# `per_m_history` and `integrator_meta` for rotating-basis runs, which the
+# point file does not carry. So compare key sets — top level and one level
+# into `dynamics` — and leave both files alone if the point file is missing
+# anything. Keys only; no frame is read, so this costs milliseconds.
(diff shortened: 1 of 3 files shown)


anko9801:
Parked as draft — anko notes the `result.jld2` / `point_NNN.jld2` duplication is **intentional**, `result.jld2` being a summary artifact. So the premise of this PR (that it is unintended) is wrong, and collapsing it to a symlink would change the design rather than fix a bug. Not merging.

One measurement worth leaving on the record: as written, `result.jld2` is **9.71 GB against point_001's 10.25 GB** — it carries all 753 frames, so it is not functioning as a summary size-wise. If the summary role is what is wanted, dropping `dynamics/psi_snapshots_streamed` from it would serve that role *and* recover the space, which is a different and better change than this one. Left for a decision rather than done.

The gate `test/workflow/test_result_jld2_dedupe.jl` and the `_COST` entry are independent of that call and could be salvaged if useful.

PR #196, opened 2026-07-30
docs(spgpe): retract the euv3 result — prescribing mu prescribes N0

anko9801:
anko, on the previous PR: *"if you fit it a certain way BEC forms or doesn't depending on the setting? that doesn't represent real physics, does it?"* — correct, and the objection reaches further than the fit. This PR retracts the claim and records why.

## The verdict tracked a one-parameter fit

$\mu = \mu_{\rm TF}(N_0)$ follows the 0-D condensate down as $K_3$ eats it, so:

| $K_3$/fit | 0-D final $N_0$ | vs measured | $\mu$ peak@ | $G_{\rm eff}$ | verdict |
|---|---|---|---|---|---|
| 1 | 1789 | **0.04** | 7.4 % | 4.5 | SHORT |
| 0.5 | 3630 | 0.07 | 8.8 % | 8.5 | marginal |
| 0.25 | 7200 | 0.14 | 10.4 % | 15.3 | FORMS |
| 0.02 | $6.1\times10^4$ | 1.21 | 19.5 % | 60.8 | FORMS |

The fitted $K_3$ leaves **4 %** of the measured $5.02\times10^4$ (PRL 129, 223401) — a 28× shortfall, and $\mu\sim N_0^{2/5}$ makes that $\mu$ under-estimated ~3.8×. Reproducing the measurement takes $K_3\approx$ fit/50, close to the independent BEC-fit estimate rather than to the value in use, and there the budget clears by 8.8×. The verdict flips at $K_3/{\rm fit}\approx0.3$.

**The $K_3={\rm fit}/50$ run was deliberately not made.** Tuning the input until the desired answer appears is the thing being objected to.

## The deeper problem is the ensemble, not the fit

I proposed taking $\mu$ from the I region instead, so $N_0$ would be an output — and measured it before believing it:

| branch | $\mu$ range | drive $\mu-\varepsilon_0>0$ at |
|---|---|---|
| `:condensate` | $-4.46\to1.74$ (max 10.80) | 436/447 points |
| `:thermal` | $-4.46\to0.00$ (max **0.00**) | **0/447 points** |

An ideal Bose gas caps $\mu$ at 0 while the trap ground state is $\varepsilon_0=\tfrac32\bar\omega=0.62$–$1.50$. The drive is negative *everywhere*: $N_0\equiv0$ is **imposed**, not predicted. My proposal was wrong; it is now documented as a trap rather than offered as the principled option.

The real statement: in a **grand-canonical** SPGPE, $\mu$ below $\varepsilon_0$ forbids a condensate and $\mu$ above it *sets* the equilibrium size via $\mu=\varepsilon_0+c_0n_0$. **Prescribing $\mu$ is prescribing $N_0$**, whatever $\mu$ is derived from. No choice of source fixes it.

## What survives

Stated plainly in the guide now, and narrower than before:

- **The lag**, for a *given* $(T(t),\mu(t))$. $\mu$ peaks 95 of 1288 internal units in (0.053 s of a 0.714 s window) and 77 % of the budget lands after the turnover where damping *removes* condensate. That the 0-D quasi-static assumption breaks at the handoff is a real finding.
- **That the solver condenses when the reservoir allows it** — 82 % of $N_{\rm TF}$ at fixed $\mu=5$, $T=2$, still rising.
- The parameter-free implementation gates (Rooney's published coefficients, RJ total to 4 %, number conservation to $10^{-13}$, Eq. 29 to 5 %, GPU=CPU, condensation to $N_{\rm TF}$), which were never in question.

## What it takes to answer the atom-number question

A **number-conserving** formulation, with $\mu(t)$ solved so that $N_C+N_I$ matches a measured total. Not implemented; recorded as the gap rather than papered over.


anko9801:
## The fast tier has never reported on this PR, and it fails

Every CI run on this branch ended `cancelled` — four of them (30733987166, 30734825588, 30735136197, 30735350520). The rollup therefore read `SUCCESS,CANCELLED,SUCCESS,SUCCESS,SUCCESS,SKIPPED` and looked like a scheduling nuisance. It was not: **`Test (Julia 1.12, tier=fast)` is genuinely red**, and the cancellations kept that invisible.

Run locally instead, on this branch merged with current `main` (`32d2b0c6`):

```
SPGPE equilibrium atom number vs the Rayleigh-Jeans mode sum:
  Test Failed at test/dynamics/test_spgpe_equilibrium_number.jl:125
  Expression: rj.condensed

  c0_run   = 0.19
  N_run    = 150819.15
  N_TF     =  78947.37
  ratio_TF = 1.9103759060908674
  rj_applies = true
```

The other two cells pass (`c0 = 0.02` → ratio 0.876, `c0 = 0.002` → ratio 0.960, both `rj_applies = false`). Only the `c0 = 0.19` cell, the one where the Rayleigh-Jeans criterion is declared to apply, comes out at **1.91× the Thomas-Fermi number**.

The failure aborts the file, so the runner then reports `unrun files (tier=fast)` — a second error that is a consequence, not an independent one.

**It is not a merge interaction.** Re-run on `8e3c99b3` with no `main` merged at all, the failure is bit-identical: same cell, same `ratio_TF = 1.9103759060908674`, same line. This PR is red on its own.

Left open and not merged. The other six PRs in this batch (#236, #237, #240, #245, #200, #288/#294) are on `main` with green CI against the base they actually landed on.

anko9801:
## Re-checked on `cc160ac8`: same failure, and the test contradicts its own prose

Two commits landed after my first report (`a2ab5e26`, `cc160ac8`). Neither touches this test or the SPGPE source — only `docs/guides/figures/kz_toroidal_winding.jl` — and the failure is **bit-identical** across three independent trees:

| tree | `ratio_TF` at `c0 = 0.19` |
|---|---|
| `8e3c99b3`, no `main` merged | 1.9103759060908674 |
| `8e3c99b3` + `main` | 1.9103759060908674 |
| `cc160ac8` + `main` (`32d2b0c6`) | 1.9103759060908674 |

So it is neither a merge interaction nor stale.

### The assertion asks for the opposite of what the comment predicts

```julia
# Weak coupling at mu = 15 is deeply CONDENSED, so the Thomas-Fermi number
# mu*V/c0 is the prediction there and Rayleigh-Jeans does not apply at all.
for c0_run in (0.19, 0.02, 0.002)
    ...
    @test rj.condensed          # all three are condensed
    @test r.N ≈ N_TF rtol=1.0
end
```

The prose explains condensation as a **weak**-coupling property, then the loop demands it at the **strongest** coupling as well. The run agrees with the prose:

| `c0` | `ratio_TF` | `rj_applies` |
|---:|---:|---|
| **0.19** | **1.910** | **true** — i.e. NOT condensed |
| 0.02 | 0.876 | false |
| 0.002 | 0.960 | false |

`d = k²/2 + 2c₀n − μ`, so a larger `c₀` lifts `d` through the mean-field shift and the `d ≤ 0` condensate signal never fires. `c0 = 0.19` is the least-condensed cell of the three, which is also why it is the only one far from Thomas-Fermi.

### A second thing worth a look

`@test r.N ≈ N_TF rtol=1.0` admits anything within a factor of two, so it passes at 1.91 without objecting. Whatever happens to the `condensed` assertion, that tolerance is close to vacuous where it matters.

I have not changed either — which of the two is wrong (the assertion, or the comment and the cell list) is a physics call, not a merge call.

anko9801:
## Pushed: `condensed` is now the selector, and the strong-coupling cell is recorded as open

`@test rj.condensed` was asserted of all three cells and `c0 = 0.19` does not satisfy it. That is the red, and it is not a merge artefact — reproduced bit-identically on this branch alone, on this branch + `main`, and on `cc160ac8` + `main` (`ratio_TF = 1.9103759060908674` every time).

### The measurement that changed my mind mid-fix

My first plan was "assert each cell against whichever predictor applies." Then I printed the Rayleigh-Jeans number at `c0 = 0.19`:

| | run | prediction | ratio |
|---|---:|---:|---:|
| Rayleigh-Jeans mode sum | 150819 | 65936 | **2.29×** |
| Thomas-Fermi | 150819 | 78947 | 1.91× |

**It matches neither.** So there is no way to write that cell green, and the plan was wrong. The header's question — *"A factor of 6.7, and one of the two is wrong"* — is still open at the strong-coupling end, and nothing in this commit answers it.

### What landed

- **`condensed` becomes the selector, not a blanket expectation.** The file's own prose already describes condensation as a *weak*-coupling property; `0.19` is the strongest of the three. Raising `c0` lifts `d = k²/2 + 2c₀n − μ` through the mean-field shift, so the `d ≤ 0` signal never fires. That cell is the least condensed, which is also why it is the one far from Thomas-Fermi.
- **Condensed cells → Thomas-Fermi, `rtol` 1.0 → 0.15.** Measured 0.876 and 0.960. The old bound admitted anything inside a factor of two, which is exactly how it passed the 1.91 without objecting.
- **`c0 = 0.19` → `@test_broken`**, not deleted and not given a loose bound. The discrepancy stays visible with both ratios in the comment, and the suite goes **red the moment it closes** — so whoever resolves it is told rather than left to notice.
- **The trailing fourth `run_equilibrium(0.002)` is dropped.** The loop now asserts exactly that at the same tolerance; it was a full extra equilibrium solve.

Local run of the file: no `Fail` and no `Error` column, one `Broken` — the one added here.

**This is a recorded open discrepancy, not a resolved one.** If you would rather the cell be dropped than marked broken, or if 2.29× points at something in the projector convention you already suspect, say so and I will change it.

anko9801:
## The CI red was a 15-minute budget, not a test failure — and my first two explanations were wrong

Three per-PR runs died at **15m18s, 15m21s, 15m20s**. `.github/workflows/ci.yml:129` sets `timeout-minutes: 15` on the fast job, and `##[error]The operation was canceled` is what that prints.

I got there badly, so both wrong turns are worth recording:

1. I read `gh pr checks` reporting `fail` and concluded the test suite was red on merit. The log said `cancelled`.
2. I then blamed pushes to this branch recomputing `refs/pull/196/merge`, re-ran in a window where **nothing** pushed — and it died at 15m20s again. Three near-identical durations is a budget, not a race. I should have read that off the first two.

### The measurement

| file | time |
|---|---:|
| **`dynamics/test_spgpe_equilibrium_number.jl`** | **1266 s (21.1 min)** |
| `dynamics/test_spgpe.jl` | 96.9 s |
| `dynamics/test_thermal_cfield.jl` | 2.1 s |
| `analysis/test_vortex_counter_control.jl` | 2.0 s |
| `analysis/test_coherence_length.jl` | 1.4 s |

One file, three SPGPE equilibrium solves at `n = 48`, each run to steady state. `main`'s fast tier is 6–10 min, so the headroom was about five minutes and this needs twenty-one.

**Worker count cannot help.** The fast job is parallelised, but a parallel makespan is bounded below by the slowest single file — a 21-minute file blows a 15-minute budget at any width.

### Pushed

- **Moved to `FULL_EXTRA`.** It is the only tier whose budget (30 min) it fits, and it takes two thirds of that. That is an argument for making the test cheaper at the source, not an endorsement of the move — three solves at `n = 48` is where the time is.
- **Registered `_COST = 1266.0`.** It had no entry, so the balancer used `_DEFAULT_COST = 3.0` and handed out **last** the one file that sets the makespan. Measured on a 10-core box rather than the runner; the ordering it buys does not depend on that, since nothing else in any tier is within a factor of four.
- The other two files this branch adds to `FAST_TESTS` are 2.1 s and 1.4 s and stay.

`test_tier_membership.jl` passes, cost-model checks 111 → 112.

### Still open, and not mine to close

- **The `c0 = 0.19` discrepancy** from the previous comment: the run matches neither convention (2.29× Rayleigh-Jeans, 1.91× Thomas-Fermi), marked `@test_broken`.
- **The nightly `full` workflow is already failing** (08-01 23:22 and 17:19), independently of this branch. Worth knowing before relying on `full` to run the file I just moved there.

PR #198, opened 2026-07-30
test: judge tests by what they catch — inventory, mutation harness, and the four defects they found

anko9801:
## Why

The suite reached 326 files and ~2,800 s of modelled serial cost by depositing one
file per incident, with nothing recording what any file defended — so nothing could
ever be removed, and the failures were dominated by tests defending a spelling or a
fitted number rather than physics. This PR adds the two instruments that make
"necessary / redundant" measurable, writes down what a test here is allowed to
claim, and fixes the four defects the instruments found on their first run.

`docs/conventions/testing_strategy.md` is the authority document.

## What a test may claim

Without an experiment, correctness can only come from five things: `exact`,
`order`, `invariant`, `metamorphic`, `differential`. A numeric literal from a past
run (`pin`) and a key name (`api`) detect *change*, which is often a bug, but
cannot support a physics claim — and they are the tests that go red on every
legitimate refactor.

Two further rules the incident record forces:

- **Coverage is (claim × path), not claim.** `ws.lhy` has been dropped on six
  separate paths while the sign was single-declared and correct.
- **A claim is tested at the lowest layer that can express it.** A defect caught
  at L4 arrives with no localisation and costs orders of magnitude more.

## Instruments

**`test/_inventory.jl`** — classifies every test file by the strongest thing it can
prove, and by layer. Keyword-based, so it over-credits: a reading order, not
evidence.

> 46 files primary-labelled `pin`, 23 `api`, 6 unclassified — 23 % cannot ground
> anything. 117 files classified L0 carry 847 s, i.e. **389 ms per assertion**;
> nothing at that price is a primitive test.

**`test/mutation/`** — the incident log in executable form: 15 defect classes that
actually happened here, each reduced to the smallest textual edit that reproduces
its shape, anchored by a regex that must match exactly once (a refactor that moves
the site fails the run rather than quietly testing nothing). Nightly and
on-demand, never a PR gate — one package precompile per mutant. Restores through
an on-disk sidecar so a killed run cannot leave a mutant in `src/`.

> First results, 12 classes × 65 oracle files: a **7-file, 32 s subset catches
> every class anything catches**. 47 of 65 caught none of the 12.
> `zeeman_quadratic_sign` is defended by exactly one file — and by none of the
> three named for Zeeman. `trap_omega_not_squared` likewise, because almost every
> fixture traps at ω = 1 where the mutation is the identity.

## Defects found

| | |
|---|---|
| `_COMP_BLANES_MOAN_S6` is **4th order, not 6th** | Blanes & Moan SRKN₆ᵇ — the 6 counts stages. Documented as "the default 6th" with Yoshida-6 as the "~15 % slower" alternative, so anyone choosing it for high order silently got 4th. Measured 4.00 three ways: a generic non-commuting pair, the (kinetic, potential) split the propagator performs, and a pair satisfying the RKN condition `[B,[B,[B,A]]] = 0` exactly. **BREAKING**: `:blanes_moan_s6` → `:blanes_moan_srkn6b`; no `runs/` config selects it. |
| `test_term_consistency.jl` covered **5 of 14 terms** | Its header says "for every term in the HamTerm registry". Doubling the kinetic energy left it green. Now sweeps the registry (9 terms FD-checked) with the terms this fixture cannot activate listed and asserted, so a term can be uncovered but not uncovered-and-invisible. |
| `lhy: {kind: scalar}` was a **silent no-op without DDI** | The c_lhy auto-derivation was guarded on `c_dd > 0`. Scalar LHY is the plain LHY correction of a contact gas — most meaningful *without* DDI. ε_dd now enters only when the DDI is actually in the Hamiltonian, and an underivable coefficient warns instead of vanishing. |
| `column_density_movie` threw `UndefVarError` | Its returned NamedTuple referenced `sources`, a local of a different function. `multi_step: true` path only. |

The dt < 0 spin-mixing no-op that collapsed Yoshida-4 to 2nd order (bisected to
#183) was fixed on main independently by 03dc152c, a Taylor-Horner performance
rewrite whose message does not mention dt < 0 — the defect was removed as a side
effect, by an author with no reason to know it was there. The merge takes main's
version; the gate stays, because nothing else pins the claim.

## New gates

- **`test/oracles/test_negative_dt_substeps.jl`** (FAST) — no substep may be the
  identity on the dt < 0 half-line, over the whole registry plus exact
  reversibility of the spin-mixing substep at frozen mean field.
- **`test/hamiltonian/test_composer_order_conditions.jl`** (FAST, 2.1 s) — composer
  order from the coefficients on 8×8 matrices, no grid and no Workspace, against
  the 14 s Eu F=6 6³ DDI order test that lives in `full` and never runs on the PR
  gate.
- **`test/workflow/test_yaml_physics_reaches_workspace.jl`** (CI) — the
  (block × path) table. A new path is a row, not a new file.

Every one is **canaried**: run against a deliberately broken version and required
to go red. That is not ceremony — three of the fixtures written for this PR passed
against the defect they were written to catch (F = 1 dispatches to a different
kernel; a Gaussian tail does not reach the guard's 1e-14 floor; a single-sign c₁
fixture misses half of a signed-product guard).

## Also

`run_chunk.jl` writes a per-file PASS/FAIL verdict and `runtests.jl` names the red
FILES — worker stdout is buffered until the worker exits, so a red parallel run
previously named the worker, not the test.

## Verification

`ci` tier green on the merge result (268 files, 8 workers). `test_yoshida_ddi_order.jl` 5/5.

An earlier `ci` run reported `✗ 1 file(s) never completed` with all workers exiting
0 and no test failures — the signature of a `claim()` hole where a failed `close`
returned "not mine" after the claim file existed, silently skipping the item. Fixed
in d7c8517b. It did not reproduce, so the causal link is not established; the hole
is a defect either way.

## Not in this PR

`test_combined_spin_step.jl` (red since before #183), the two F32 mixed-precision
tolerances, the L-BFGS gradient-floor pin, `pulse_sequence`'s `:zeeman` / `:B`
mismatch, `B.theta_deg` having no YAML path, and `test_quality.jl`'s
ExplicitImports. Consolidating the `workflow/` files the (block × path) table now
covers waits on a mutation run with a `workflow/` probe.


anko9801:
## Second batch: the `full`-tier red list, triaged by measurement

`full` on the merge result was 7 files red. Four are now fixed, one turned out to
be a real defect that this PR deliberately leaves red, and two are still open.

### Fixed

| | |
|---|---|
| **F32 kinetic energy was garbage** | `EnergyContext.fft_buf` was declared `Array{ComplexF64, ND}`; the buffer goes to plans built for ψ's eltype, and FFTW's in-place plans reinterpret memory. On a Float32 workspace the kinetic term came back **0.0035 against 0.48** — 99.3 % missing, total ~10 % low, every other term agreeing to 1e-8. Fixed at all six sites that hand a buffer to a plan. The two `test_mixed_precision*` thresholds were reading a real defect, not a fitted number: both runs were also unconverged at `n_steps=300`, but running each to convergence left the same 8.8 % gap. |
| **`pulse_sequence` B overlay discarded** | `haskey(compiled, :B)` where the compiler writes `:zeeman` — always false, so the `TimeDependentZeeman` was thrown away. Measured: waveform slots all `nothing` before, `RampWaveform` after. No `runs/` config uses it. `apply: zeeman` and `apply: trap` were also silent no-ops and now throw; `trap` leaves the docs. |
| **Negative elapsed time** | `t_itp = -0.322 s`: `time()` is not monotonic (NTP, WSL2 resume). `elapsed_s(t0_ns)` states the rule once and twelve sites use it. `runtime_seconds` is written to disk and the autopilot budget totals GPU-hours from it, so this is accounting, not a progress line. |
| **`test_quality.jl` / `test_combined_spin_step.jl`** | Both were wrong about correct code — see the commit messages. `ExplicitImports` cannot see macro-expansion uses (deleting the "stale" import breaks the whole CUDA extension); and the two propagators implement different quadratic-Zeeman conventions **by design**, which `_rtp_use_combined_step` already documents and guards, so the test was asserting agreement in the one regime the design excludes. |

### Left red on purpose

`solvers/test_lbfgs_accuracy_floor.jl` — "Newton-CG polish tightens the gradient
floor". Measured at this commit, same problem, separate processes:

```
ratio polish/lbfgs = 0.156          (polish engaged, ψ moved 2.6e-10)
                     0.081  ×3      (engaged, reproducible within a process)
                     1.000          (NO-OP: grad_norms bit-identical)
                     1.998          (engaged and made the residual WORSE — the full-tier failure)
```

The absolute floor also moves ~10× process to process while the energy is stable
to 1e-16, so the residual is round-off-limited and plan-dependent. But
`1.000` and `1.998` are not scatter: whether the polish runs at all, and whether
it helps, is unstable. That is a defect in `newton_polish`, not in the test, and
memory already records a bit-identical (no-op) polish on clean main. Widening the
tolerance would hide it, so the test stays red and flagged.

### Still open

`oracles/test_spin_chain_fusion_parity.jl` (GPU bit-identity) — likely no longer
achievable now that the fused and unfused arms can select different rotation
realizations (`SPIN_TAYLOR_ENABLED`), which would make it a claim to restate
rather than a bug. Not yet measured.

Verification: `ci` green on the merge result, 268 files across 8 workers.

PR #204, opened 2026-07-30
Close the convention diff against Matsui et al.'s published code, and pin Fig. 4B as a type-C target

anko9801:
Session **S-A6**. Worktree `/home/suzume/workspace/BEC-simulation-matsui`, branch `research/matsui-convention-diff`.

`docs/validation/parameter_contract_with_Ueda.md` had sat fully unfilled since 2026-05 waiting on a channel to the Ueda lab that never opened. Zenodo 17303925 (CC-BY-4.0) ships a third-party implementation of the same Hamiltonian, so §0 fills the table against that instead. **Convention independence and reference generation — not physics independence.** Same model, same Ref-(19) lineage. It does not replace S-A1 / S-A2.

## Every §0.3 row is now filled or marked absent

The two that decide whether any comparison is meaningful:

**Row 2.3 — MATCH, at the operator level.** Their `H ⊃ +ZeemanP·F_z` with `ZeemanP = +g_F μ_B B/(ħω)`; ours is `−p·F_z` with `p ≡ −g_F μ_B B/(ħω_ref)`. Same operator; the intermediate `p` differs in sign only because we route through Kawaguchi-Ueda's `p` and they do not. `inspect` returns `p = −153.9` at their `Bfield_ini`; theirs is `+153.87`. Row 2.3's own text was **stale** and contradicted `Units.bfield_to_p` — said out loud now. The smoke confirmed it end to end: ground state converges to `Mz = −6.0`, m=−6 at 100 %, under a **positive** `Bz`.

**Rows 4.1–4.5 — MATCH in the product.** Their kernel is our `Q` times `4π`; their `cdd` absorbs `μ₀/4π`. `cdd·DD ≡ c_dd·Q` identically, including `μ = g_F μ_B` per spin matrix. Neither half is "the" convention. Only the `k=0` bin genuinely differs (theirs `−4π/3`, ours `0`), worth **5×10⁻³ nT** at their box — analytic bound, §0.3.3.

Absent in their code, so our knob has no counterpart: **secular DDI** (row 4.7), dealiasing, LHY, spin-rotating frame, thermal seed (`noise_amp` is declared and never read). On the secular advisory specifically: `ω_L/(c_dd⟨n⟩) ≈ 373` was printed at the *start* of the ramp; the field is **held** 400× lower, so during the physics the ratio is ≈ 0.93. Their use of full MDDI is not evidence against CLAUDE.md's `> 100` — it is a run that never enters that regime.

Also: component ordering is ascending `m` against our descending (`c = F+1−m`), `∫|ψ|² = 1` on both sides, Condon-Shortley ladder identical, and `aHO = √(ħ/2mω)` puts a √2 in every length and coupling.

## The shipped code is provably not the published configuration

`Ntot = 3.5×10⁴` against theory curves totalling **49999.9**; and `initial.f90` ships `cc0_eff = 1, cc1_eff = 0` against `time.f90`'s `0.5 / 50`, so as shipped the state handed to the dynamics is the ground state of a Hamiltonian with **twice** the contact repulsion — `R_TF` +15 %, peak density −34 %. Peak density sets the dipole field, which sets the offset Fig. 4B measures. Parameters have to be reconstructed; `fig4b_gsvariant_n32.yaml` prices the ambiguity instead of assuming it away.

## Fig. 4B: the target, measured rather than quoted

| curve | dip centre | half-depth width |
|---|---|---|
| their **simulation** | **−2.5495 nT** | **15.0224 nT** |
| their **experiment** (4/3 repeats averaged) | **−3.2048 nT** | **14.5414 nT** |

The paper contrasts an *analytic* −1.5 nT with −3.5 nT observed. Neither is the number to reproduce: **their own code lands at −2.55 nT**, 70 % of the way across, and that is what a second implementation of the same model has to hit.

New `resonance_dip` (generic: parabolic vertex, half-depth crossings against a **per-side** baseline, because a dip on an asymmetric background has no single top). Gated in tier `ci` by `test/validation/test_matsui_fig4_dip.jl`, 35 assertions, including a canary that fails if the metric stops tracking the curve, and the mutual consistency of the two sheets (Fig. 2C at 5 ms = 21449; Fig. 4B interpolated to +2.6 nT = 21344, 0.5 %).

Fixtures: CSV extracts of all four `.xlsx` sheets under `test/fixtures/matsui2025/` with CC-BY-4.0 attribution. The `.xlsx` originals and the 1.81 GB `dataset_raw.zip` stay out of the repo.

## Runs — submitted, not yet reported

UGE array `8304456` (tasks 1-4) + ci tier `8304457`, from commit `a78a5680`, clean tree, ancestor-gated against all 14 refs in `fix_list.toml`. **No numeric result from these is claimed in this PR.** Read §0.6 for what will and will not be quotable: 32³ violates the charter's ≥4-points-per-healing-length guard (and so does their own production run at 0.7), which is why the 64³ spot check is a task and not an afterthought.

The smoke earned its keep twice: it caught `populations`, which is not an analyzer name, and it measured the step at **27 ms at 32³ D=13 with the padded DDI** — ~70× the campaign cost table's extrapolation, which resized the scan from 81 fields to 45.

## Claim labels

| claim | type | commit | tier |
|---|---|---|---|
| Fig. 4B dip centre / width of both published curves | C | `bbbe4829` | `ci` |
| Fig. 2C early-time landmarks; two-sheet consistency | C | `a78a5680` | `ci` |
| `k=0` kernel bin worth 5×10⁻³ nT | A (analytic) | `bbbe4829` | — |
| `ω_L/(c_dd⟨n⟩) ≈ 0.93` at the held field | A (arithmetic on a logged value) | `4e2fb85d` | — |
| `c1_ratio = 1/36` ⇒ `c₀ = 2343.63`, matching their `cc0_eff = 0.5` | A | `bbbe4829` | — |
| SpinorBEC's own Fig. 4B centre / width | — | **not measured yet** | — |


anko9801:
## Results — UGE 8304651, commit `f7433171`, clean tree, all four tasks exit 0

Full write-up in `docs/validation/parameter_contract_with_Ueda.md` §0.7.

### Fig. 4B

| | centre [nT] | half-depth width [nT] |
|---|---|---|
| **SpinorBEC 32³** (45 fields, −13 … +9 nT @ 0.5) | **−2.138** | **14.62** |
| Matsui simulation, same window, same metric | −2.5495 | 12.7524 |
| difference | **+0.41** | **+1.87 (+15 %)** |
| Matsui experiment (full window) | −3.2048 | 14.5414 |

**The sign and the scale of the resonant-EdH offset reproduce** from an independent implementation with independently-declared conventions. The offset is **16 % short**, the width **15 % wide**.

**It is not the grid.** Five fields re-run at 64³ agree with 32³ to **≤ 0.06 %** on `N_{−6}` (0.010 / 0.022 / 0.033 / 0.004 / 0.055 %) across a 2× refinement on independently built grids, FFT plans and padded DDI kernels. The transfer is driven by the bulk mean field, and the bulk is resolved at 32³. It is **not the kernel conventions** either — §0.3.3 bounds the `k=0` bin and the truncation difference at ~0.02 nT combined, 20× too small.

### Fig. 2C — partially, and only early

| landmark | SpinorBEC | Matsui | ratio |
|---|---|---|---|
| t at `N_{−6}` = 90 % | 0.940 ms | 1.129 ms | 0.83 |
| t at 70 % | 1.838 ms | 2.199 ms | 0.84 |
| t at 50 % | 2.908 ms | 3.588 ms | 0.81 |
| `N_{−6}` at 5 ms | 0.2721 | 0.4290 | 0.63 |
| `N_{−6}` at 40 ms | 0.6334 | 0.3618 | 1.75 |

The phenomenon reproduces, but **our transfer runs ~20 % fast** — all three landmarks at 0.81–0.84, a near-constant *rate* ratio, not a shape difference — compounding to a 37 % gap by 5 ms. Past ~10 ms the curves are unrelated, which is expected: that regime is finite-trap revivals.

**A ~20 % rate excess and a ~16 % offset shortfall are the same size, and both point at the effective dipolar drive.** That is the next thing to chase. No convention difference found in §0.3 accounts for it.

### Task 3 is VOID, and that is the useful part

`c1_ratio: 0.0` in a `ground_state` step should double `c₀` (2343.63 → 4687.27, verified at the function level) and **does not reach the run**: the two arms' converged ground-state energies are **−910.66795 vs −910.66985**, 2×10⁻⁶ apart, where ~30 % was due. Their dip centres differ by 0.02 nT.

Two arms that should differ and are numerically indistinguishable is evidence about the **plumbing**. The §0.3.5 ground-state ambiguity stays **unpriced**, and the fix needs a gate that fails when the value does not move.

### Two defects found and worked around, both worth their own fix

1. **`run_yaml` writes the ground state as `point_001`'s top-level `psi`** in every multi-point scan — 45/45, 19/19 and 6/6 point scans each had exactly that one file wrong; a single-point run had none. In a B-scan the first point is an endpoint, which is where a dip's baseline comes from: reading it inflated the measured width from 12.7 to 16.6 nT before I caught it. The report now reads the dynamics series and refuses if its last sample is not at the end of the hold.
2. **`SPINORBEC_TEST_WORKERS=auto` OOM-kills a 16-core node** (UGE 8304457, exit 137 at 201 s) — one julia per core, each loading SpinorBEC into a clean session by design.

### Retraction

The **27 ms/step** figure I published earlier is wrong and is retracted in `9ba8c800`. It came from an ITP progress timer over 200 smoke steps still dominated by first-call compilation. Measured warm: **5.2 s per 32³ point (~1.5 ms/step)** and **21.7 s per 64³ point (~6.3 ms/step)**. The campaign cost table is ~4× optimistic at 32³, not 70×. It cost the scan 36 fields it did not need to lose.

### Still open

`ci`-tier job 8304652 is in flight; no green-tier evidence yet beyond the new file passing standalone (35/35).

anko9801:
## Correction to the framing above — and the `ci` tier is green

### I was treating their curve as the reference. It isn't one.

The 0.41 nT centre gap was written up as *our* shortfall. That assumes they are right, which is not established. Their implementation has three properties that could each bias it — a 2nd-order finite-difference Laplacian at **0.7 points per healing length**, the internally inconsistent shipped ground state (§0.3.5), and an `Ntot` in the shipped code that is not the one behind the published curves. Ours has an under-resolved grid and a ground-state ambiguity we **failed to price**, because task 3 is void.

**The experiment cannot arbitrate the centre.** The Fig. 4 caption states the field axis "may contain an offset error of up to 10 nT" — **three times the dip centre itself**. So −2.14 (ours), −2.55 (theirs), the measured −3.20, and even the paper's own −1.5 nT analytic estimate are all consistent with that axis. Any claim that a simulated centre "agrees with experiment" here is claiming more than the measurement supports — including the paper's own.

### The width can arbitrate, and it does not favour us

An axis offset shifts a dip; it does not stretch one. On the identical −13 … +9 nT window:

| | half-depth width [nT] | vs experiment |
|---|---|---|
| Matsui experiment | 12.84 | — |
| Matsui simulation | 13.07 | **+1.8 %** |
| SpinorBEC 32³ | 14.62 | **+14 %** |

On the one Fig. 4B quantity immune to their stated systematic, **their simulation tracks their measurement an order of magnitude better than ours does**. One number and one observable is weak evidence, but it points at our side, not theirs. The paper reads the width as the spread of the dipole field over the cloud, so a 14 % wide dip says our field distribution is broader than the real one — which sits consistently beside the ~20 % fast transfer in Fig. 2C.

(The width is baseline-sensitive at the third digit: the same curve gives 13.07 over [−13, +9] with 42 points and 12.75 when the window edge falls one point differently. Two significant figures, not four.)

**So the order of work is to close our own unknowns first** — the `c1_ratio` plumbing defect and the ground-state ambiguity it left unpriced, which §0.3.5 puts at 34 % of the peak density — before attributing any of the gap to them.

Written up in `docs/validation/parameter_contract_with_Ueda.md` §0.7 at `0b645e6c`.

### `ci` tier

**Green** at `11b8a83d` (UGE 8304717, exit 0, 818 s, 4 workers). The earlier red (8304652) was this PR's own file failing to load: `DelimitedFiles` is a stdlib that still has to be declared in the test target. Replaced with a direct read — the header needs a quote-aware split since its fields contain commas.

anko9801:
## 実測に寄せた版 — 結果と所要時間 (UGE 8304841, `0e78456e`, 両タスク exit 0)

### 1. 実測との最大の差は「原子数が保存していない」こと

Published experimental sheet summed at 5 ms: the total runs from **45.8k in the wings to 27.6k at the dip** — 8 % to 45 % below 5×10⁴, and **deepest exactly where the dip is**. Fig. 4B plots absolute `N_{−6}`, so that field-dependent deficit sits inside the compared quantity, and both simulations conserve atom number.

### 2. 三体損失では説明できない — 135 点で測定

| `K₃` [cm⁶/s] | atom loss at 5 ms (−12.5 … +9 nT) | dip centre | width |
|---|---|---|---|
| 1.2×10⁻²⁹ (Miyazawa 2021 direct) | 0.27 – 0.44 % | −2.146 | 14.63 |
| 3.6×10⁻²⁹ (3×) | 0.80 – 1.31 % | −2.161 | 14.08 |
| 1.2×10⁻²⁸ (10×, above anything measured) | 2.61 – 4.25 % | −2.213 | 14.07 |
| **measurement** | **8 – 45 %** | −3.205 | 12.84 |

Ten times the larger published `K₃` gets to 4 % against a measured 45 %, an order of magnitude short, and its field dependence is flat where the measurement's is peaked at the dip. **Condensate three-body loss is not what the deficit is.** Left to look at: detection across 13 Stern-Gerlach spots (harder exactly when the population spreads, i.e. at the dip), thermal fraction, shot-to-shot preparation.

The dip barely notices: 10× `K₃` moves the centre 0.07 nT and the width 0.6 nT. **Adding loss does not reconcile our curve with theirs.**

### 3. 無料の補正2つ

- **1 nT field jitter** (stated in the caption, = a Gaussian average over B): shifts either centre by 0.05 nT. Width moves ≲0.6 nT and in opposite directions on the two curves — an edge artefact of the truncated kernel against an endpoint baseline, not a result. It does not close a 1.4 nT width gap.
- **Normalising the atom-number deficit away** changes the measured dip very little: centre −3.205 → −3.139, width 12.84 → 12.41. So the deficit is not what separates experiment from simulation *in shape*.

### 4. `c1_ratio` の欠陥を確定 — 上の数値はすべて暫定

Three ground states, scanning only `c1_ratio` over 1/36, 0, 0.5 — a **19× range** in `c₀` (2343.63 / 4687.27 / 246.70):

| `c1_ratio` | `c₀` due | converged `E` |
|---|---|---|
| 1/36 | 2343.63 | −910.6679 |
| 0 | 4687.27 | −910.6698 |
| 0.5 | 246.70 | −910.6692 |

Span **1.3×10⁻³** where ~8.6 was due, and **not monotonic in `c₀`**. The non-Zeeman part of that energy is ≈12.6 against a non-interacting 1.59, so the contact term dominates it — not a small effect hiding under a large one. `load_config` resolves the step dict correctly, so the loss is downstream of parsing.

**Consequence:** the dynamics step inherits `ws_prev.interactions`, so it got whatever the ground state got. Contract rows 3.2/3.3 assert our `c1_ratio = 1/36` reproduces their `cc0_eff = 0.5 / cc1_eff = 50` pair exactly — **verified at the function level, not verified for what ran**. An unintended `c₀` with `c₁ = 0` instead of `c₀/36` is a live candidate for both the 20 % rate excess and the 15 % width excess, so **every Fig. 4B / Fig. 2C number is provisional until this is fixed and the scan re-run.**

### 5. 所要時間 — 予測と実測

| | predicted | measured |
|---|---|---|
| task 5 (135 pts, loss on, 32³) | 13 min | **1210 s = 20.2 min** |
| task 6 (3 ground states) | < 2 min | **311 s = 5.2 min** |

Per-point, warm: **5.2 s** loss-free at 32³, **6.2 s** with loss (+19 %), **21.7 s** at 64³. First point of each task absorbs ~80–107 s of JIT.

**The lesson for the cost model: whole-job wall is ~1.5× the sum of per-point times.** Predicting from per-point cost alone under-called this job by 1.55×; the rest is JIT plus per-point workspace rebuild, analyze and I/O. Predict from walls. (This is the second time this session a cost figure was wrong — the first was the retracted 27 ms/step.)

Reference for the whole set: a 45-field 32³ scan is **~10 min**, a 135-point one **~20 min**, a 6-point 64³ one **~7 min**.

anko9801:
## 訂正 — 「5 ms で 8〜45 % の原子数減少」は損失ではありません

I read the Fig. 4B sheet's total-number deficit as something a loss model might have to reproduce. That was wrong, and half the atoms cannot go in 5 ms on any reading. **Their own data settles it**, in a sheet already committed as a fixture:

`dataset_figs3.csv` is their trap-lifetime measurement, 4 repeats:

- `N(0) = 54396`, `N(1.6 s)/N(0) = 0.532` ⇒ **1/e time 2.54 s** ⇒ **0.20 % loss in 5 ms**.
- Even treating the whole 0 → 1.6 s drop as a linear rate — an over-estimate for a decaying-rate curve — gives 80 atoms, **0.15 %**.
- 45 % in 5 ms would need an **8.4 ms** lifetime, **303× faster than they measured themselves**.

**Our published-`K₃` arm was right all along**: 0.27–0.44 % against their S3's 0.20 %. There was never a 45 % for a loss model to reproduce.

### What the deficit actually is

It tracks **how far the population has spread across the Stern-Gerlach spots**. Over the 61 measured fields, correlation(reported total, number of components carrying > 2 %) = **−0.935**:

| B [nT] | reported total | components > 2 % | columns reported negative |
|---|---|---|---|
| −17.50 | 41954 | 4 | 2 |
| −8.17 | 31406 | 9 | 2 |
| −5.83 | **27187** | **10** | 1 |
| +8.17 | 45039 | 3 | 3 |
| +17.50 | 45595 | 2 | 5 |

Ten faint, broad spots instead of two bright ones and each fit under-counts; several columns come back **negative**, which is background subtraction, not physics. So Fig. 4B compares an absolute `N_{−6}` carrying a field-correlated **counting** bias against two simulations carrying none.

It does not rescue our gap either way: normalising it away moves the measured dip very little (centre −3.205 → −3.139, width 12.84 → 12.41), because the bias is largely common to numerator and total. And since `m = −6` is the brightest spot it is under-counted *least*, so the fraction is if anything biased *up* at the dip — the direction that makes the measured dip shallower than truth, not deeper.

### What the K₃ ladder was still worth

Over a **10× range** in `K₃` the dip centre moves **0.07 nT** and the width **0.6 nT**. So loss cannot be tuned to close the 0.41 nT / 1.4 nT gap with their curve either — which is a real exclusion, just not the one I set out to test.

Docs corrected at `3dad17b0` + `a6fde573` (the first of those dropped the K₃ table with an over-wide replacement; the second restores it).

anko9801:
## 撤回 — `c1_ratio` にバグはありません。私のプローブが縮退方向を振っていました

I reported a confirmed plumbing defect. It is not one. The missing column was `c₀ + 36c₁`:

| `c1_ratio` | `c₀` | `c₁` | `c₀ + 36c₁` | converged `E` |
|---|---|---|---|---|
| 1/36 | 2343.633 | 65.101 | **4687.2663** | −910.6679 |
| 0 | 4687.266 | 0 | **4687.2663** | −910.6698 |
| 0.5 | 246.698 | 123.349 | **4687.2663** | −910.6692 |
| −0.005 | 5716.178 | −28.581 | **4687.2663** | — |

A fully polarised `m = −F` state has `|⟨F⟩| = F n`, so its contact energy density is `(c₀/2)n² + (c₁/2)F²n² = ((c₀ + F²c₁)/2)n²` — **only the combination enters**. And `interaction_params_from_constraint` holds that sum fixed, because that is the `c₀ + 36c₁ = 4π(a_s/a_ho)N` constraint doing exactly its job.

**So the ground-state energy is *supposed* to be independent of `c1_ratio`.** The probe scanned a direction the observable is degenerate in, and had no positive arm — nothing in it would have moved even with perfect plumbing, so "it did not move" carried zero information. UGE 8304841 task 6 stands; only my reading of it was wrong.

### 三つの帰結

1. **The Fig. 4B and Fig. 2C numbers are NOT provisional for this reason.** Rows 3.2/3.3 stand as verified. Please disregard the "every number is provisional" paragraph in the earlier comment.
2. **§0.3.5's ground-state ambiguity does not exist.** Their `initial.f90` (`cc0_eff = 1, cc1_eff = 0`) and `time.f90` (`0.5 / 50`) both give `c₀ + 36c₁ = 8πN a₀/aHO`, so **their polarised ground state is the same state under either setting**, and so is ours. The difference bites only once the spin depolarises — i.e. in the dynamics, where they use `0.5 / 50`, which is the `c1_ratio = 1/36` we match. The only real remaining item in §0.3.5 is `Ntot`. (Residual: `initial.f90` runs its ground state unpadded, `sn = 1`.)
3. **The 20 % rate excess and the 15 % width excess lose their leading candidate explanation and are now unexplained.**

### 教訓

This is the campaign charter's positive-control rule, and I broke it in the direction that is easy to miss: a null result from a control that could not have been non-null. A does-the-knob-reach-the-solver test needs an observable **not degenerate in the knob** — for `c1_ratio` that means a state with `|⟨F⟩| < F n`, or a direct read of `ws.interactions`, not a polarised ground-state energy. `gs_c1ratio_probe.yaml` is kept with that written into it.

Retracted at `096adc14`.

anko9801:
## 三点まとめて完了 (UGE 8307358, `87cc221d`, 両タスク exit 0)

### (b) 64³ full 45-field scan — **32³ は収束していました**

| | centre [nT] | width [nT] |
|---|---|---|
| SpinorBEC 32³ | −2.138 | 13.97 |
| **SpinorBEC 64³** | **−2.145** | **13.96** |
| Matsui simulation, [−12.5, +9] | −2.549 | 12.75 |

Centre 0.007 nT apart, width 0.01 nT apart, worst-case point-wise difference in `N_{−6}` over 44 fields **0.48 %** (typically < 0.1 %). **The 0.40 nT centre gap and 1.21 nT width gap are not resolution.** 1297 s.

(The earlier 6-field 64³ arm is superseded — it was chosen for a point-by-point check and cannot bracket a dip; `resonance_dip` on it returns −5.32 nT, an artefact of extrapolating a vertex across a 7.5 nT gap.)

### (a) Their exponential ramp — **not the explanation**

Built exactly: `B(t) = (B_ini − B_fin)e^{−t/τ} + B_fin`, τ = 50 µs, as a constant (the scan axis) plus a piecewise-linear exponential tail at τ/6. `inspect` confirms the composition (`bz` starts at −151.9 = B_ini + B_fin rather than −153.9).

The hypothesis was worth testing — their ramp needs 347 µs to fall within 10 nT where ours was done at 150 µs, so we had a ~200 µs head start out of 5000, and a 200 µs shift would have put their 90 % landmark at 0.929 ms against our measured 0.940.

**It moves the centre 0.07 nT of the 0.41** (−2.138 → −2.208) **and leaves the width alone** (13.97 → 14.02). Worth using because it is their actual protocol; not what separates the curves.

### `point_001` は報告より悪い — 幅の数値を訂正します

Its dynamics series carries the **last** scan point's result:

| scan | `N_{−6}` at −13 nT (pt 001) | at +9 nT (pt 045) |
|---|---|---|
| `fig4b_scan_n32` | 36404.886940 | 36404.886940 |
| `fig4b_theirramp_n32` | 36704.978256 | 36704.978256 |

Ten digits at two fields whose published values differ by 13 %. The **centre is untouched** (a vertex through the minimum's neighbours, 20 points from either endpoint); the **width is not**, since its left half-depth level rides the endpoint baseline.

**Our width excess is +9.6 %, not the +14 % I reported.** Every width in the document is now the point-1-dropped number.

### 残ったもの

| candidate | verdict |
|---|---|
| grid resolution | **excluded** — 64³ vs 32³, 0.48 % worst case |
| ramp shape | **excluded** — 0.07 nT of 0.41 |
| three-body loss | **excluded** — 10× `K₃` worth 0.07 nT; real loss is 0.2 % |
| kernel conventions | **excluded** — ~0.02 nT |
| ground-state ambiguity | **does not exist** (retracted above) |
| 1 nT field jitter | **excluded** — 0.05 nT |
| the experiment's atom-number deficit | **not loss** — a counting systematic |

What survives is a **uniform ~20 % excess in atoms moved out of `m = −6`** — ratio ours/theirs = **0.78–0.84 at every field across the dip**, and the same 20 % as the Fig. 2C rate excess. That is one number in the coupling, not a protocol difference. Two candidates remain: something in the effective dipolar drive we have not found, and their own discretisation (2nd-order FD at 0.7 points per healing length, against our converged spectral result). The width evidence still tilts toward the first.

### (c) Merge

CI green. Merging this now; the remaining ~20 % is a fresh lane item.

PR #205, opened 2026-07-30
Migrate the five icosahedral arms to full_bdg, and label the polar arms' ansatz mismatch

anko9801:
Follow-up to #203, which made the F=6 I_h closed form refuse `λ_spin < 0`. Five
configs could no longer run; this migrates them and reports what looking at them
turned up.

## The migration, measured rather than assumed

At the ladder these suites actually use — `c1_ratio: -0.005`, `N_atoms: 30000`,
Eu151 F=6:

```
c_0 = 2681.44   λ_spin = -187.70   λ_spin/c_0 = -0.070   ⇒  icosahedral: NaN
full_bdg, same ladder: table builds, all values finite, V_LHY(n=3.7e-3) = 0.2303
```

So they move to `kind: full_bdg`, the path the guard names, and the files are
renamed to match — a file called `LHY_icosahedral.yaml` carrying `full_bdg` would
be the naming lie the convention forbids. `cli.jl inspect` on the migrated configs
plus two siblings: parse clean, no `:block` / `:error` / `:warn`.

There was no generator to fix instead. The headers name
`scripts/validation/eu_k3_lhy_gen.jl`, `eu_k3_lhy_control_gen.jl` and
`eu_lhy_longtime_gen.jl` — **none of the three exists in the tree**, nor does the
"scope note" they refer to. Recorded in the annotations.

## What the migration exposed

All **twelve** arms across these three suites are `initial_state: m_minus_F`, an
FM state. The closures applied to them are `icosahedral` (I_h ansatz),
`polar_contact` (polar ansatz, ζ = δ_{m,0}) and `scalar`. Only `scalar` — the
fully-polarised approximation — matches the state. The guard exposed the
icosahedral half by making it error; **the polar half computes a number from the
wrong ansatz and says nothing.**

The polar arms are deliberately **not** rewritten. `runs/eu_k3_lhy/*.yaml` label
this "applied outside nominal regime", so the mismatch is this family's study
design rather than an accident, and replacing it would change what an
already-void suite measures. Each polar arm now carries a note that its number is
not a validated closure for this state, naming `fm_contact` as the matching closed
form and `full_bdg` as the general path — so the comparison is not read as three
valid closures.

## Does polar/fm need the same guard? No — but FM has a different one

`scripts/probe_lhy_closed_form_refusal_shapes.jl` measures all three.

**polar is not the same defect.** `kappa_m = abs(delta_m)` is correct: δ is an
anomalous coupling and ω² = ξ² − |δ|² depends on |δ|. The instability surfaces as
`t = ξ/κ − 1 < 0`, and `phi_1_reg` **regularises** it, saturating at
φ₁^reg(−1) = 0.3177 — the documented Petrov droplet limit. That is the opposite of
masking. Driving σ_0 negative (reachable at g_0 = −20) raises a bare
`DomainError`: less informative than the I_h refusal, but not silent.

**FM has a different defect of the same class — found, not fixed.**
`lhy_energy_fm`'s `kappa < 1e-12 && return 0.0` was meant as "negligible" and also
swallows negative `g_{2F}`, so a negative stiffness silently reports **no LHY**:

| c1/c0 | g_12 | ε_fm |
|---:|---:|---:|
| −0.005 | 8.2 | 10.4048 |
| −0.0278 | −0.008 | **0.0** |
| −0.05 | −8.0 | **0.0** |

Reachable at F=6 whenever `c1/c0 < −1/36 = −0.0278`, since `g_{2F} = c0 + 36 c1`.
Left alone on purpose: unlike the I_h guard this behaviour change has not been
measured against the runs that depend on it, and silent-zero LHY is exactly the
failure mode six other paths have already produced.

## A correction to my own probe

Its first version asserted "σ_0 < 0" while σ_0 was in fact **positive** (0.846 at
g_0 = −1) and read a healthy number as a pass. It now prints the stiffness it
claims to test, so the label cannot drift from the condition.

## Verification

`cli.jl inspect` on 5 configs and the `full_bdg` build at the production ladder,
TSUBAME jobs 8304437 / 8304465 (cpu_4). Config-only plus one probe script — no
source change, so #203's test evidence still stands.

The full `ci` tier is running separately against merged main (`10fb9121`, job
8304428) to close the gap #203 left open; that result is about main, not this
branch, and is reported when it lands.

Type A (code correctness) throughout.

### DIFF
--- a/runs/eu_k3_lhy/LHY_full_bdg.yaml
+++ b/runs/eu_k3_lhy/LHY_full_bdg.yaml
@@ -1,13 +1,21 @@
 # Task #19C — LHY interference at K3=200× cigar regime.
 # Source: scripts/validation/eu_k3_lhy_gen.jl
-# LHY kind = icosahedral (applied outside nominal regime; see scope note).
+# LHY kind = full_bdg (general-spinor BdG; valid for this state and ladder).
+# Was `kind: icosahedral`, migrated 2026-07-30. The I_h closed form now
+# refuses this ladder: c1_ratio = -0.005 gives lambda_spin/c_0 = -0.07, i.e. a
+# negative spin-Goldstone stiffness, where it used to take |lambda_spin|^(5/2)
+# and overshoot full_bdg by up to 11%. `full_bdg` is the general-spinor path
+# the guard directs callers to and it is valid here. NOTE the state is
+# `m_minus_F`, so the matching CLOSED form would be `fm_contact` (any F since
+# 2026-07-27, gated against full_bdg at F=1..8) at ~100x less cost; full_bdg is
+# chosen to keep this suite's arms distinct from a future fm_contact arm.
 
 metadata:
   suite: eu_k3_lhy
   ladder_level: 12
   parent_regime: cigar_N30k_omz0p25
   K3_factor: 200.0
-  LHY_kind: icosahedral
+  LHY_kind: full_bdg
   generator: scripts/validation/eu_k3_lhy_gen.jl
 
 defaults:
@@ -27,7 +35,7 @@ pipeline:
       ddi:
         enabled: true
         secular: false
-      lhy: {kind: icosahedral}
+      lhy: {kind: full_bdg}
       B: {Bz: "0.01 Gauss", theta: 0.0, phi: 0.0}
       gauge_fix: false
       initial_state: m_minus_F
@@ -45,7 +53,7 @@ pipeline:
         phi: 0.0
       ddi:
         secular: false
-      lhy: {kind: icosahedral}
+      lhy: {kind: full_bdg}
       loss:
         K3_per_m_si:
           ["2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s"]
--- a/runs/eu_k3_lhy/LHY_polar_contact.yaml
+++ b/runs/eu_k3_lhy/LHY_polar_contact.yaml
@@ -1,6 +1,19 @@
 # Task #19C — LHY interference at K3=200× cigar regime.
 # Source: scripts/validation/eu_k3_lhy_gen.jl
 # LHY kind = polar_contact (applied outside nominal regime; see scope note).
+#
+# ANSATZ MISMATCH, left in place deliberately (noted 2026-07-30). This arm's
+# state is `m_minus_F`, i.e. FM, while `polar_contact` is the closed form for
+# the polar state (zeta = delta_{m,0}). runs/eu_k3_lhy/*.yaml label this
+# "applied outside nominal regime", so the mismatch is the study design across
+# this family, not an accident — but the number it produces is NOT a validated
+# closure for this state and must not be read as one. The matching closed
+# form is `fm_contact` (any F since 2026-07-27, gated against full_bdg at
+# F=1..8); the general path is `full_bdg`. The sibling `icosahedral` arm had
+# the same defect and could no longer run at all once the I_h form began
+# refusing lambda_spin < 0, so it moved to full_bdg. This arm fails QUIETLY
+# instead, which is the more dangerous half. Neither the "scope note" nor any
+# generator named in these headers exists in the tree.
 
 metadata:
   suite: eu_k3_lhy
--- a/runs/eu_k3_lhy_control/LHY_full_bdg_K0.yaml
+++ b/runs/eu_k3_lhy_control/LHY_full_bdg_K0.yaml
@@ -1,7 +1,15 @@
 # K3=0 LHY model control — auto-generated.
 # Source: scripts/validation/eu_k3_lhy_control_gen.jl
-# LHY = icosahedral, K3 = 0, γ_dr = 0.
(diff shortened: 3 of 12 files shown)


anko9801:
Two things from #201/#206, which measured the same regime from the `full_bdg` side. Both are cross-cutting with #209 rather than objections to this PR.

## 1. This PR and #209 do the same migration by incompatible means

They touch **disjoint filenames**, so git will not flag a conflict — but they are the same five configs:

- **here**: `rename from runs/eu_k3_lhy/LHY_icosahedral.yaml` → `LHY_full_bdg.yaml` (×5)
- **#209**: edits `runs/eu_k3_lhy/LHY_icosahedral.yaml` **in place** to `kind: full_bdg`, keeping the name

Merge both and the tree ends up with `LHY_full_bdg_K0.yaml` *and* an `LHY_icosahedral_K0.yaml` that contains `full_bdg` — duplicated suites, and the second one is precisely the *"a file called `LHY_icosahedral.yaml` carrying `full_bdg` would be the naming lie the convention forbids"* that this PR's own body rules out. Whichever lands second should drop its half of the migration.

## 2. `finite` is not the same as `defined` — measured at this PR's own ladder

The migration rests on *"full_bdg, same ladder: table builds, all values finite, V_LHY(n=3.7e-3) = 0.2303"*. Finite, yes. But at that ladder `full_bdg` **reports the mean field dynamically unstable**:

```
c0 = 3270.05   c1 = -16.3502    (c1_ratio = -0.005, N_atoms = 30000, B = 0.01 G)

state        V_LHY(n=3.7e-3)   finite   unstable   max Im ω
m_minus_F         0.59327       true      true       0.396
polar             0.644428      true      true       0.893
```

Its own warning is what that means:

> the zero-point sum drops the complex branches while the counterterms still subtract all 13 of them, so ε_LHY is **scheme-dependent** here. This is a property of the state — **the closed forms are no better**.

So the migration makes these configs *runnable* — which is a real improvement over erroring at table build — but not *quotable*. That is worth saying in the headers, because "moved to the general-purpose path" reads as "now valid".

On the texture campaign (a different ladder: N=50000, c₁ ∈ {0, 0.018}, 50–80 µG) the same instability holds at every cell, seed and converged, and none of the escape routes the warning names is reachable — q is 15 orders of magnitude too small, c₀/c₁ are Eu's scattering lengths, and converging buys a factor 4–13 and stops. Full record: `docs/validation/full_bdg_scheme_dependence_eu_f6.md` (#201), with the three-mode table in #206.

## Where this PR is right, and it matters

The observation that **all twelve arms are `initial_state: m_minus_F`** while the closures assume I_h / polar ansätze is independently confirmed here: `m_minus_F` and `polar` give max Im ω of 0.396 vs 0.893 at identical couplings, so the state genuinely changes the answer. Labelling the polar arms rather than rewriting them is the right call for an already-void suite.

anko9801:
Superseded by the follow-up branch: #207 landed the same config migration in parallel, so the overlapping half of this PR is redundant. The parts that were not duplicated — the FM silent-zero, the hole it slips through in #209's new domain gate, the tier submitter's BLAS pin, and the polar arms' ansatz-mismatch note — move to a branch based on current main.

anko9801:
**Correcting my comment above — my recommendation of "not full_bdg because it's scheme-dependent" was right about full_bdg and wrong about the conclusion.** One more measurement changes the target.

## The instability is entirely dipolar

Switch the DDI off:

| seed | max Im ω, DDI on | max Im ω, DDI off |
|---|---:|---:|
| m_minus_F (this ladder) | 0.396 | **0** |
| flower (texture ladder) | 0.365 | **0** |
| polar_core_vortex | 0.926 | **0** |
| polar | 1.030 | **0** |

Exactly zero. The contact problem is perfectly stable; the dipolar term at `ε_dd = 0.5402` is the whole instability. And with the DDI off, `full_bdg` and `fm_contact` agree to six significant figures — `V(n=3.7e-3)` = 0.380007 vs 0.380006 — cross-validating both in the stable regime.

## So `fm_dipolar` is the correct migration target, not `full_bdg`

Two prescriptions for the same physics, and only one is scheme-dependent:

- **`full_bdg`** drops the complex branches from the zero-point sum while the counterterms still subtract all `D` of them — its own definition of scheme-dependent.
- **`fm_dipolar`** uses Petrov's prescription: `lima_pelster_Q5` zeros the integrand where `1 + ε_dd(3cos²θ − 1) < 0`, excluding the unstable angles rather than half-counting them. `ε_dd = 0.54 < 1` is inside that domain (guard documented at `fm_dipolar.jl:26`), and it returns a finite **unwarned** 0.553422 — checked, not merely silent, which is the distinction I had to verify before claiming it.

And this PR already points at it: *"naming `fm_contact` as the matching closed form"*. For these suites, which have DDI enabled, `fm_dipolar` is the dipolar member of that pair. The ansatz matches for the reason this PR identified — all twelve arms are `m_minus_F`, and `c₁ < 0` makes FM the mean-field ground state.

```
m_minus_F, c1_ratio=-0.005, N=30000, DDI on:
  fm_contact   V=0.380006   finite  unwarned
  fm_dipolar   V=0.553422   finite  unwarned
  full_bdg     V=0.59327    finite  UNSTABLE  (max Im ω = 0.396)
```

## Suggested change to this PR

Keep the rename — file name = content is the convention, and #209's in-place edit leaves the naming lie you correctly rule out — but **rename to `LHY_fm_dipolar*.yaml` with `kind: fm_dipolar`**, rather than `full_bdg`. That is a valid domain rather than a runnable one, so the headers can say the number *is* usable instead of carrying a caveat.

The `polar_contact` arms are a separate matter and I agree with labelling rather than rewriting; if you ever do rewrite them, `polar_dipolar` is their dipolar counterpart, since contact-only closures cannot see the term that is the entire instability here.

Corrected record: `docs/validation/full_bdg_scheme_dependence_eu_f6.md` (#206). My earlier "no usable LHY mode in this regime" was an overreach — it was missing the two `*_dipolar` rows.

PR #208, opened 2026-07-30
bench: a phase-gap accuracy instrument, and why an LHY reference cannot be tabulated from a seed

anko9801:
An instrument for "how much accuracy does separating two ground-state phases
need", and the answer to why it cannot be run the way I was running it.

## The question, and why it has no single number

On a phase boundary the competing states are degenerate by construction, so
ΔE → 0 and no finite accuracy resolves the boundary exactly. The answerable
question is how far the boundary MOVES:

    δB_boundary ≈ δ(ΔE) / |∂(ΔE)/∂B|

which makes the observable the **gap between two competing states**, not the
total energy — errors common to both cancel, and only the non-cancelling part
decides a phase. `SPIN_TAYLOR_TOL` was budgeted against the total energy
(2.4e-13 out of E ≈ 10), which licenses an energy claim and says nothing about a
boundary.

Note the direction that helps, because it is not the obvious one: a **small**
`|∂(ΔE)/∂B|` is the bad case. It means the phases stay near-degenerate over a
wide range of B, so any energy error moves the apparent boundary a long way — and
far enough that way, what is there is a crossover and not a boundary at all. No
amount of accuracy fixes that; only a different claim does.

## The finding: `full_bdg` cannot be tabulated from a seed

The reference arm never converged — `final dE` stuck at 3e-3…1.1e-2 against a
`tol` of 1e-9, while the `polar_contact` arms reached 1e-5…1e-7. The reason was in
the log all along:

```
FullBdG LHY: mean field is dynamically unstable (max Im ω = 1040.0).
The zero-point sum drops the complex branches while …
```

`FullBdGLHY` warns exactly when the mean field is dynamically unstable, and
CLAUDE.md records that ε_LHY is **scheme-dependent** there. So the reference arm
had no well-defined ground state to converge to, and the non-convergence was
correct behaviour rather than a numerical failure.

The root cause is structural: **`:full_bdg` tabulates from the peak-density
spinor of the state the workspace is built with.** Starting ITP from
`:flower` / `:chiral_spin_vortex` means the table is built from an *unrelaxed*
seed — dynamically unstable — and then used for the entire relaxation. The
reference was scheme-dependent by construction.

**So this is a methodological requirement, not a bug and not issue #172:** an
LHY reference has to be tabulated from a RELAXED state. Two stages — relax with
no LHY or a closed form, rebuild the workspace with `full_bdg` from the relaxed
ψ, re-converge. That applies to every use of `full_bdg` as a reference, not just
to this measurement.

I reported "not converged" and then "blocked on #172" before getting here. Both
were wrong, and the evidence that would have settled it was in the first run's
output.

## What the instrument does

`bench/phase_gap_error_budget.jl` — four arms (reference / production / dt/2
baseline / operator-removed control) across a small B scan, reporting ΔE,
convergence **with the final dE value** (a boolean cannot distinguish "not
converged" from "converged, criterion never fired"), the per-component **winding
vector** that classifies the state, and the state separation.

It refuses to produce a number unless both seeds converged AND landed in
different winding classes. Across three runs it refused three times, for three
different correct reasons: unconverged arms; its degeneracy guard firing (two
seeds relaxing to the same state, whose difference measures nothing — this
project has read exactly that as physics once); and now dynamic instability.
Refusing is the design.

Also uses `tol_drho`, the gauge-invariant gate — a texture carries a global U(1)
phase and a spin rotation about z that leave ρ alone while keeping dE/|E| moving.

## Process fixes, and why they are in this PR

Four times this session a runner discarded its own diagnosis. All four are fixed
here or in #202:

* **`grep -vE "^│|^└|^┌"` swallowed every `@warn`.** It was there for the CUDA
  library-path boxes on every node; with stderr merged into the pipe, the
  `full_bdg` instability warning reached neither `.o` nor `.e`. It hid the finding
  above for two full runs. Now filters by **content** — the specific CUDA lines —
  never by form.
* **`tail -N` cut exception headers** (#202): a depot fault read as
  `156/156 → 2 passed / 14 errored`, which is exactly what a regression in that
  PR's change would look like.
* **A smoke inside the job.** `submit_load_check.sh` checks that `src/` loads; it
  cannot see an undefined variable in a bench script. The smoke earned its cost on
  its first run, catching `!=` on a vector carrying `missing` (returns `missing`,
  not `Bool`) that would otherwise have surfaced after ten minutes of converged
  ITP. Production is now unreachable unless the tiny run passed.
* **Never edit a remote worktree in place.** A `sed` on TSUBAME left it dirty,
  `git checkout` said only `Aborting`, and the job nearly ran a stale script.

## Scope

`bench/` and `scripts/tsubame/` only — no `src/`, no `test/`. The 32³ production
arm is still running; the diagnosis above is from the smoke arm's warning and the
completed 32³ run that preceded the filter fix, and does not depend on it.


anko9801:
**Confirmed at 32³** — the production arm the PR body said was still running has finished, and the diagnosis holds at production scale rather than only in the smoke:

```
FullBdG LHY: mean field is dynamically unstable (max Im ω = 1040.0)   ← smoke, 8³
FullBdG LHY: mean field is dynamically unstable (max Im ω = 1020.0)   ← production, 32³
```

Verdict unchanged: **0 usable B points**, and the reference arm's `final dE`
sitting at 3e-3…1.1e-2 against a `tol` of 1e-9 while the `polar_contact` arms
reach 1e-5…1e-7.

One caveat on reading the warning count: it is 2 for ~24 `full_bdg` workspace
builds, i.e. one per Julia process — the warning is `maxlog`-limited. So the count
says the instability is present in each process, **not** how many of the arms or
B points it covers. Establishing that per-arm would need the limit lifted; the
conclusion here does not depend on it, since a single scheme-dependent ε_LHY in
the reference is already enough to disqualify the reference.

The `dt/2` baseline is worth a note too: its ΔE is remarkably flat across B
(−16.885, −16.904, −16.946, −16.927) where the reference swings −51 to −115. With
neither converged that is not interpretable, but it is the kind of pattern that
would be worth a second look once the two-stage tabulation is in — a baseline
that does not move with B would mean the slope, and therefore any δB, cannot be
extracted from it either.

PR #209, opened 2026-07-30
test(oracles): gate committed configs against their LHY mode's validity domain

anko9801:
## The gap

`IcosahedralLHY` returns NaN at `λ_spin < 0` — for these parameters, at `c₁ < 0`.
`runs/eu_k3_lhy*` use `c1_ratio = −0.005`, so their `icosa` cells had been
measured **outside the domain**, and that reached a committed
`factorial_2x4.json`, a committed figure and two documented claims before anyone
noticed (#207). I found it cleaning up after #199, not from a test.

Nothing was watching:

| layer | what it checks | why it missed this |
|---|---|---|
| schema | `kind` is a known string | `icosahedral` *is* known |
| `_tabulate_lhy` | NaN → `ArgumentError` | fires at **build** time, i.e. only when a run is launched — long after review and merge |
| — | the config vs the domain | **nothing** |

## The gate

Walks the configs already in the tree and evaluates each one's closed form at
**its own** `(F, c₀, c₁)`. Not a restatement of the domain rules — it calls the
same functions a run would and fails on the same NaN.

**It found five configs on its first execution**: the `icosahedral` cells in
`eu_k3_lhy/`, `eu_k3_lhy_control/` and `eu_lhy_longtime/` (×3). #207 corrected
the *artifacts* but left the configs naming an invalid mode, so a re-run would
have died at table build. All five now say `kind: full_bdg`, each with a header
explaining why so the next reader does not switch them back.

## Two guards against the gate going vacuous

- it asserts it **found configs at all** — a silent empty sweep is the exact
  failure shape this file exists to catch;
- it **canaries the domain guard itself** with `eu_k3_lhy`'s own numbers
  (`c₀ = 3270.05`, `c₁ = −16.35` must be NaN; `+16.35` must be finite). If
  `epsilon_LHY_F6_Ih` stops rejecting, every other assertion here would pass for
  the wrong reason.

## Scope

The cheap closed forms (`icosahedral`, `polar_contact`, `fm_contact`).
`full_bdg` and `spatial` cost minutes per config and are the general-purpose
paths with no ansatz to violate — they are the *answer* to a domain failure, not
a source of one.

## Verification

- fast tier: **173 files, all passed**
- oracles tier: one failure, `test_spin_chain_fusion_parity.jl` — **unrelated and
  pre-existing**. No LHY reference, and it fails identically with these changes
  stashed, under `--check-bounds=yes` (the runner's flag; it disables `@inbounds`
  and changes the SIMD reduction order the bit-exactness assertion depends on).

### DIFF
--- a/runs/eu_k3_lhy/LHY_icosahedral.yaml
+++ b/runs/eu_k3_lhy/LHY_icosahedral.yaml
@@ -1,3 +1,10 @@
+# LHY kind was `icosahedral` until 2026-07-30. `IcosahedralLHY`'s closed form
+# is c0^(5/2) + 3|lambda_spin|^(5/2); the absolute value made it symmetric under
+# c1 -> -c1 and returned a real energy where the spin-Goldstone branch is
+# dynamically unstable. This config runs at c1_ratio < 0, i.e. OUTSIDE that
+# domain -- `epsilon_LHY_F6_Ih` now returns NaN and the table build throws.
+# `full_bdg` diagonalises the coupled problem with no ansatz and is valid here.
+# Gated by test/oracles/test_lhy_config_validity_domain.jl.
 # Task #19C — LHY interference at K3=200× cigar regime.
 # Source: scripts/validation/eu_k3_lhy_gen.jl
 # LHY kind = icosahedral (applied outside nominal regime; see scope note).
@@ -7,7 +14,7 @@ metadata:
   ladder_level: 12
   parent_regime: cigar_N30k_omz0p25
   K3_factor: 200.0
-  LHY_kind: icosahedral
+  LHY_kind: full_bdg
   generator: scripts/validation/eu_k3_lhy_gen.jl
 
 defaults:
@@ -27,7 +34,7 @@ pipeline:
       ddi:
         enabled: true
         secular: false
-      lhy: {kind: icosahedral}
+      lhy: {kind: full_bdg}
       B: {Bz: "0.01 Gauss", theta: 0.0, phi: 0.0}
       gauge_fix: false
       initial_state: m_minus_F
@@ -45,7 +52,7 @@ pipeline:
         phi: 0.0
       ddi:
         secular: false
-      lhy: {kind: icosahedral}
+      lhy: {kind: full_bdg}
       loss:
         K3_per_m_si:
           ["2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s", "2.0e-39 m^6/s"]
--- a/runs/eu_k3_lhy_control/LHY_icosahedral_K0.yaml
+++ b/runs/eu_k3_lhy_control/LHY_icosahedral_K0.yaml
@@ -1,3 +1,10 @@
+# LHY kind was `icosahedral` until 2026-07-30. `IcosahedralLHY`'s closed form
+# is c0^(5/2) + 3|lambda_spin|^(5/2); the absolute value made it symmetric under
+# c1 -> -c1 and returned a real energy where the spin-Goldstone branch is
+# dynamically unstable. This config runs at c1_ratio < 0, i.e. OUTSIDE that
+# domain -- `epsilon_LHY_F6_Ih` now returns NaN and the table build throws.
+# `full_bdg` diagonalises the coupled problem with no ansatz and is valid here.
+# Gated by test/oracles/test_lhy_config_validity_domain.jl.
 # K3=0 LHY model control — auto-generated.
 # Source: scripts/validation/eu_k3_lhy_control_gen.jl
 # LHY = icosahedral, K3 = 0, γ_dr = 0.
@@ -9,7 +16,7 @@ metadata:
   ladder_level: 12
   parent_regime: cigar_N30k_omz0p25
   K3_factor: 0
-  LHY_kind: icosahedral
+  LHY_kind: full_bdg
   generator: scripts/validation/eu_k3_lhy_control_gen.jl
 
 defaults:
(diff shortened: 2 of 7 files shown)


anko9801:
The gate itself looks right to me — evaluating each config's own `(F, c₀, c₁)` through the same function a run would call, rather than restating the domain rules, is the part that makes it hard to fool, and the two anti-vacuity guards (found-configs-at-all, plus canarying `epsilon_LHY_F6_Ih` with `eu_k3_lhy`'s own numbers) are exactly what was missing elsewhere today. Two cross-cutting notes.

## 1. This PR and #205 do the same migration by incompatible means

Disjoint filenames, so git will not flag a conflict — but the same five configs:

- **here**: edits `runs/eu_k3_lhy/LHY_icosahedral.yaml` in place to `kind: full_bdg`, keeping the name
- **#205**: `rename from runs/eu_k3_lhy/LHY_icosahedral.yaml` → `LHY_full_bdg.yaml`

Merge both and the tree gets `LHY_full_bdg_K0.yaml` *and* an `LHY_icosahedral_K0.yaml` containing `full_bdg`. #205's body calls that second file *"the naming lie the convention forbids"*, so the two PRs disagree about the fix, not just its mechanics. Whichever lands second should drop its half.

## 2. The scope note is right about ansätze and incomplete about `full_bdg`

> `full_bdg` and `spatial` … are the general-purpose paths with no ansatz to violate — they are the *answer* to a domain failure, not a source of one.

True for ansatz violations. But `full_bdg` has a validity condition of its own — **mean-field stability** — and these very parameters violate it. Measured at `eu_k3_lhy`'s ladder:

```
c0 = 3270.05   c1 = -16.3502    (c1_ratio = -0.005, N_atoms = 30000, B = 0.01 G)

state        V_LHY(n=3.7e-3)   finite   unstable   max Im ω
m_minus_F         0.59327       true      true       0.396
polar             0.644428      true      true       0.893
```

Its own warning: *"the zero-point sum drops the complex branches while the counterterms still subtract all 13 of them, so ε_LHY is scheme-dependent here. This is a property of the state — the closed forms are no better."*

So the five migrated configs will now **build and run** instead of dying at table build — a real improvement — but they are not thereby in a valid domain. Same shape as the defect this gate was written for: a finite number from an invalid regime, with nothing saying so.

## Suggestion, if you want it in scope

The instability check is a **runtime** condition (it needs the spinor and the density), so it cannot be a pure config-level assertion like the closed-form domains. But it is reachable the same way this gate reaches the closed forms: build the `full_bdg` table at each config's `(F, c₀, c₁, q)` and its declared `initial_state`, and fail — or warn with the value — when `max Im ω` is non-zero. That is ~seconds per config at a small grid, since the BdG solve does not need the production resolution.

It would have to be a separate tier from this one (this gate is fast; that one is not), and it is a bigger claim, so it may belong as a follow-up rather than here.

Context for both: `docs/validation/full_bdg_scheme_dependence_eu_f6.md` (#201) has the texture-campaign measurements — every cell unstable, seed and converged, with all three escape routes the warning names closed. #206 adds the three-mode table, i.e. the statement that **no LHY mode is valid in this regime**, which is the sentence neither the `icosahedral` half nor the `full_bdg` half says on its own.

anko9801:
**Correcting my comment above — my recommendation of "not full_bdg because it's scheme-dependent" was right about full_bdg and wrong about the conclusion.** One more measurement changes the target.

## The instability is entirely dipolar

Switch the DDI off:

| seed | max Im ω, DDI on | max Im ω, DDI off |
|---|---:|---:|
| m_minus_F (this ladder) | 0.396 | **0** |
| flower (texture ladder) | 0.365 | **0** |
| polar_core_vortex | 0.926 | **0** |
| polar | 1.030 | **0** |

Exactly zero. The contact problem is perfectly stable; the dipolar term at `ε_dd = 0.5402` is the whole instability. And with the DDI off, `full_bdg` and `fm_contact` agree to six significant figures — `V(n=3.7e-3)` = 0.380007 vs 0.380006 — cross-validating both in the stable regime.

## So `fm_dipolar` is the correct migration target, not `full_bdg`

Two prescriptions for the same physics, and only one is scheme-dependent:

- **`full_bdg`** drops the complex branches from the zero-point sum while the counterterms still subtract all `D` of them — its own definition of scheme-dependent.
- **`fm_dipolar`** uses Petrov's prescription: `lima_pelster_Q5` zeros the integrand where `1 + ε_dd(3cos²θ − 1) < 0`, excluding the unstable angles rather than half-counting them. `ε_dd = 0.54 < 1` is inside that domain (guard documented at `fm_dipolar.jl:26`), and it returns a finite **unwarned** 0.553422 — checked, not merely silent, which is the distinction I had to verify before claiming it.

And this PR already points at it: *"naming `fm_contact` as the matching closed form"*. For these suites, which have DDI enabled, `fm_dipolar` is the dipolar member of that pair. The ansatz matches for the reason this PR identified — all twelve arms are `m_minus_F`, and `c₁ < 0` makes FM the mean-field ground state.

```
m_minus_F, c1_ratio=-0.005, N=30000, DDI on:
  fm_contact   V=0.380006   finite  unwarned
  fm_dipolar   V=0.553422   finite  unwarned
  full_bdg     V=0.59327    finite  UNSTABLE  (max Im ω = 0.396)
```

## Suggested change to this PR

Keep the rename — file name = content is the convention, and #209's in-place edit leaves the naming lie you correctly rule out — but **rename to `LHY_fm_dipolar*.yaml` with `kind: fm_dipolar`**, rather than `full_bdg`. That is a valid domain rather than a runnable one, so the headers can say the number *is* usable instead of carrying a caveat.

The `polar_contact` arms are a separate matter and I agree with labelling rather than rewriting; if you ever do rewrite them, `polar_dipolar` is their dipolar counterpart, since contact-only closures cannot see the term that is the entire instability here.

Corrected record: `docs/validation/full_bdg_scheme_dependence_eu_f6.md` (#206). My earlier "no usable LHY mode in this regime" was an overreach — it was missing the two `*_dipolar` rows.

For this PR specifically: the gate's scope already covers `fm_contact`, so extending it to the `*_dipolar` pair would keep the migrated configs inside it — whereas migrating them to `full_bdg` moves them **out** of the gate's reach, which is the opposite of what you want after finding them this way.

PR #210, opened 2026-07-30
perf(lbfgs): stop when the line search says there is no step left, and retract the earlier numbers

anko9801:
Follow-up to #194. It contains a correction to #194's headline, a change worth two orders of magnitude more than #194 was, and the removal of one thing #194 added that measurement says is harmful.

## The correction first

**#194's numbers were wrong and are retracted.** It claimed the two-loop recursion is 55–71 % of a CPU iteration and this work is 1.6–2.4× faster. Measured properly — 5 minutes per point instead of ~1 s — the two-loop is **4 %**, and an iteration is **95 % line search**.

Three compounding harness defects produced those numbers. All three are written up in `docs/design/lbfgs_iteration_cost.md` rather than quietly fixed, because they are more useful to carry forward than the numbers were:

1. per-iteration cost was `(t(20 steps) − t(4 steps))/16` — a difference of two ~1 s runs, each of which plans FFTs under `FFTW.MEASURE`, so it was largely plan-measurement noise divided by sixteen. It moved the same baseline cell 20 % between jobs;
2. `n_hi = 20` never exceeded `m_lbfgs = 20`, so no differenced step ever had a full L-BFGS history — which is how the breakdown could show a component at **>100 %** of the whole without anyone noticing;
3. the bench passed `tol = 0.0` so that `n_steps` is consumed exactly. The consequence is that essentially every measured iteration was a **post-convergence** one.

The figure and CSV are deleted with them. A stale figure is worse than none.

## Where the cost actually is

Eu-151 F=6, 24³, `node_q`, ~5 min/point:

```
per-iteration                     235.6 ms
  energy_gradient!                 12.3 ms   ( 5.2 %)
  project x2                        1.3 ms   ( 0.5 %)
  sobolev precond                   2.6 ms   ( 1.1 %)
  two-loop dir                      9.8 ms   ( 4.2 %)
  line search TOTAL               223.9 ms   (95.0 %)   29.41 evals/iteration
```

`iteration = 26 ms + n · 7.6 ms`, and it reconciles to 5 % at the measured `n`. `n` is now **measured and returned** (`n_line_search_evals`) — the bench used to solve the residual *for* `n`, which makes a breakdown close by construction and so can never report that it does not.

`n ≈ 30` is the backtracking cap. The line search was exhausting it and returning `α = 0` on **94 % of steps**. Not a solver defect: the problem's gradient floor is **5.03e-7** and the default `tol` is **1e-8**, fifty times below what an energy-gated line search can resolve. The solve reaches the floor in a few dozen steps and spends every remaining step proving it cannot move — ~472 s of wall for ~0.9 s of useful work on a default-tolerance call.

## `stop_at_floor` (default on)

A line search that fails along the **steepest-descent** direction is conclusive: it leaves ψ, `grad` and `E` untouched with an empty curvature history, so the next iteration rebuilds that exact direction from that exact gradient with that exact `E0` and repeats the same evaluations to the same answer. The iterate is a fixed point of the loop. **No count and no tolerance** — one such failure ends the solve. A failure along an L-BFGS direction is deliberately *not* conclusive: the reset to steepest descent is a genuinely different attempt and often succeeds.

An earlier draft used `max_line_search_failures = 3`; pushing the argument through showed no constant is needed and that it fires earlier.

## `floor_limited`

`converged = false` cannot distinguish a solve that failed from one that got as far as the method allows, and the second is the common case — every Eu F=6 ground state at the default tolerance has been reported as not converged. When the solver stops at its floor and that floor is above the `tol` asked for, it now says so, with both numbers and the remedy (`residual_polish=true` is not energy-gated and does reach below the floor — this was already in the driver's docstring and reached nobody).

`converged` keeps meaning `grad_norm < tol`: a lot of code reads it and it is persisted. The pipeline stores `stop_reason` and `floor_limited` beside it.

## Threading removed

#194 threaded the two-loop's axpys. Measured on an exclusive node at 24³, that is **negative**:

| threads | 1 | 4 | 16 | 48 |
|---|---|---|---|---|
| two-loop, ms | 9.8 | 22.6 | 13.2 | 32.4 |
| iteration, ms | 48.7 | 72.6 | 47.3 | 76.3 |

Each axpy is ~0.5 ms of real work with `2m` per direction, so a `@threads` region per axpy is mostly launch cost, and that cost grows with the thread count. `_axpy!` is now a sequential `@simd` loop, bit-identical to the broadcast.

`_realdot` keeps its 64-block structure for **accuracy** (one level of pairwise summation; gated against a `BigFloat` reference on an ill-conditioned input) and because it avoids BLAS `zdotc`, whose thread team is sized from the machine. Worth 2.4 % — BLAS level-1 only appears in the ~8 % of the iteration that is not the line search. Pinning BLAS process-wide would reach the same 2.4 % and also throttle the genuine level-3 work elsewhere (dense `eigen` in the Bogoliubov solver).

A negative result kept in the doc: the first `_realdot` used hand-unrolled scalar accumulators for machine-independence and measured **8.20 ms against `zdotc`'s 7.72 ms** *with its blocks threaded*. "Blocked and threaded" sounds like it must be faster and was not.

## Gates

- `stop_at_floor` gives up nothing: the stopping and grinding runs must agree on energy, `grad_norm` and ψ **exactly** — unconstrained, under `target_magnetization` (whose retraction iterates), and on GPU (whose energy goes through the fused device pass, not the CPU registry).
- The GPU gate asserts the **mechanism**, not a step budget. Two earlier versions asserted "a stall occurs within N steps" and failed because the problem had not reached its floor yet on that backend — while every assertion that mattered passed. A gate that legislates a step count is a gate whose step count gets tuned until it goes green.
- `floor_limited`'s gate derives the attainable tolerance **from the measured floor**: ask for ten times the floor the solver just reported and the same solve converges.
- `test_lbfgs_sobolev_preconditioner.jl`'s `r.converged || r.last_step >= 100` read `last_step` as a proxy for effort; it now states its intent directly. The remaining `last_step` readers are upper bounds (which only get safer) or compare two runs that move together.

## Verification

- `tier=ci` at `92fc0b96`: **262 files, all passed** (TSUBAME job 8304864).
- `tier=full` in flight; will report below. On the previous tip its failure set was a strict subset of `main`'s.

## Not in this PR

`_grad_ddi!` ignores padding while `_ddi_energy` honours it, so since `ddi_padding` defaulted on a DDI ground state descends an unpadded gradient toward a padded energy. Correctness, not performance, and it changes every padded-DDI ground state.


anko9801:
**Verification complete (TSUBAME):**

| tier | commit | result |
|---|---|---|
| `ci` | `92fc0b96` | **262 files, all passed** (job 8304864) |
| `full` | `92fc0b96` | 5/12 workers fail (job 8304865) |

The `full` failures are exactly the pre-existing set:

```
gpu/test_mixed_precision.jl:64            gpu/test_mixed_precision_phase3.jl:51
hamiltonian/test_combined_spin_step.jl:146,148
oracles/test_spin_chain_fusion_parity.jl:69
solvers/test_yoshida_ddi_order.jl:74,75,83,87
workflow/test_pipeline.jl:382
```

All ten also fail on a clean `origin/main` worktree (measured separately, job 8297857). **This branch adds none.** The new gates — the floor stop under `target_magnetization`, the GPU mechanism gate, and the `floor_limited` tolerance relationship — all pass.

Two that come and go between runs on this repo regardless of branch, for the record: `solvers/test_lbfgs_accuracy_floor.jl:118` (on `main` the Newton-CG polish returns a bit-identical ψ, so the "polish improves the floor" assertion compares two numbers at a floor that is itself not reproducible: 5.7e-10, 1.4e-8 and 2.8e-8 measured for the same commit and problem) and `workflow/test_phase_diff_eval.jl:75`.

PR #211, opened 2026-07-30
feat(validation): accuracy profiles solved from an error budget — and the measurement that says :fast is empty

anko9801:
Two named settings a config author can ask for — "as accurate as this code gets"
and "as fast as measurement licenses" — without knowing which knobs exist, which
direction each points, or what each costs. Plus the measurement that says what the
second one may contain, which turns out to be nothing.

## The speed profile is SOLVED, not chosen

The first version of this had a hand-picked `fast` value per knob, and I put
`ddi_pad_factor = 1.5` in it. **That violates the criterion in
`error_budget.jl`'s own header**: its 1.9e-2 residual Φ error is 4.4× the 4.3e-3
the production setting already carries. As long as a human picks the list, that
mistake stays available. So the list is gone.

Each knob now carries `accepted_error` — the error its *default* already has — and
a `ladder` of measured `(value, rel_error, rel_cost)` rungs.
`accuracy_profile_for_budget(frac)` returns the cheapest rung per knob whose
`rel_error ≤ frac × accepted_error`, and `:fast` **is** that solution. A knob with
no ladder cannot move at any budget, because an unmeasured trade is not a trade.

`pad 1.5` is now rejected automatically at every sane budget. Nobody decides that,
including me.

## The measurement: there is no speed in the accuracy knobs

`bench/accuracy_knob_cost.jl` — every knob, one card, one job, one production
shape. H100, Eu F=6 D=13, 32³, one ITP step against production 0.456 ms:

| setting | vs production | |
|---|---|---|
| `spinor_lhy = :polar_contact` | 0.987× | free |
| `spinor_lhy = :full_bdg` | 0.997× | free — since #202's fused diagonal took tables |
| `secular_ddi = true` | 0.986× | free, so no *speed* reason to approximate the physics |
| `ddi_pad_factor = 1.5` | **0.906×** | the only real speedup — and it fails the budget |
| `ddi_pad_factor = 3.0` | 1.507× | the price of `:reference` |
| `spin_taylor = false` | 1.613× | the price of `:reference` |
| `dealias_2_3 = true` | **1.484×** | the price of being right about aliasing — never measured before |

So the accuracy knobs are where accuracy is **paid for**, not where speed is
hiding. Today's 2.2–2.9× on the ITP step (#183/#191/#202) all came from exact
optimisations. `:fast` therefore comes out **identical to `:production`**, and the
report prints exactly that — an honest empty answer rather than a name with
nothing behind it.

Three things that follow and are worth stating separately:

* **`:reference` has a price and nobody had written it down.** It sets pad 3.0
  (1.507×) and, via the globals, Taylor off (1.613×) and dealias on (1.484×). The
  report now carries these.
* **`spinor_lhy` stopped being a speed knob today.** It was 3× before #202's fused
  diagonal accepted tabulated tables; it is 0.997× now.
* **`pad 1.5`'s speedup is grid-dependent** — 1.10× at 32³, 1.28× at 64³ — which
  is on its own an argument against any constant "fast value".

## What still needs a human

Documented in the header rather than left implicit:

1. **Only `ddi_pad_factor` has a ladder.** Every other knob is missing the error
   half, the cost half, or both. That is a gap in the data, not a property of the
   knobs.
2. **A profile sets `:per_run` knobs only.** The `:global` ones are `Ref`s; use
   `with_reference_accuracy`. The report says so rather than letting the name
   imply full coverage.
3. **`:reference` has a precondition a value cannot express** — the one that cost
   a day. `spinor_lhy = :full_bdg` tabulates ε_LHY from the peak-density spinor of
   the state the workspace is *built* with, so a raw seed tabulates from an
   unrelaxed, dynamically-unstable configuration (measured max Im ω = 1040 at Eu
   F=6) where `FullBdGLHY` itself says ε_LHY is scheme-dependent.
   `check_accuracy_preconditions` returns `:fail` for that instead of leaving it
   to a `maxlog`-limited runtime warning.

**Composition is NOT measured.** The costs are one knob at a time and they overlap
— pad and dealias both add FFTs — so a profile's cost is never inferred from its
rows.

## Gate

`test/workflow/validation/test_accuracy_profiles.jl` — 76 assertions. Pins that a
ladderless knob cannot move at any budget; that a laddered one takes the cheapest
rung within budget; that a tighter budget never picks a looser setting; and
specifically that `pad 1.5` is rejected at every sane budget **while remaining
reachable at an absurd one**, so the rejection is about the budget and not about
the rung being unreachable. The precondition check carries a positive control that
it can pass at all.

`ci` is running on TSUBAME; result to follow.

### DIFF
--- a/bench/accuracy_knob_cost.jl
+++ b/bench/accuracy_knob_cost.jl
@@ -1,0 +1,133 @@
+# What does each accuracy knob COST, on one footing?
+#
+# `ACCURACY_KNOBS` records what every knob is worth in ACCURACY, because that is
+# what a config author needs in order not to trade it away by accident. It records
+# the cost for exactly one of them. That asymmetry is why `:fast` has a single
+# member: a knob may only enter it with a measured cost AND a defensible residual,
+# and most knobs are missing the first half.
+#
+# This measures the first half for all of them, on one card, in one job, against
+# one production shape — so "which knobs are even candidates" stops being a guess.
+#
+# It does NOT measure the second half. A knob needs both, and the accuracy side
+# needs a reference and an observable (see bench/phase_gap_error_budget.jl), so a
+# cheap knob here is a candidate and nothing more.
+#
+# Two of the rows are worth stating in advance because they are not intuitive:
+#
+#   * `spin_taylor = false` is what `:reference` DOES. Its cost is the price of
+#     the reference profile, and nobody had written that down.
+#   * `dealias_2_3 = true` is the ACCURATE direction and it adds work — six
+#     complex FFTs per DDI call for the F filter, plus a ψ pre-filter. So this
+#     row is the cost of being right, not of being fast.
+#
+#   julia --project=. bench/accuracy_knob_cost.jl [n] [reps]
+
+using Printf
+import CUDA
+using SpinorBEC
+using SpinorBEC: SPIN_TAYLOR_ENABLED, DEALIAS_2_3_ENABLED, accuracy_profile
+
+include(joinpath(@__DIR__, "eu151_params.jl"))
+
+const N_GRID = length(ARGS) >= 1 ? parse(Int, ARGS[1]) : 32
+const REPS = length(ARGS) >= 2 ? parse(Int, ARGS[2]) : 30
+
+function build(; kind, pad_factor, secular, dtype=Float64)
+    grid = make_grid(GridConfig(ntuple(_ -> N_GRID, 3), ntuple(_ -> 12.0, 3)))
+    psi0 = init_psi(grid, SpinSystem(6); state=:spin_coherent,
+        init_theta=π / 4, init_phi=0.3)
+    ws = make_workspace(;
+        grid, atom=Eu151, interactions=eu_interaction_params(0.05),
+        zeeman=ZeemanParams(EU_p_weak, 0.0),
+        potential=HarmonicTrap((1.0, 1.0, EU_λ_z)),
+        sim_params=SimParams(; dt=0.002, n_steps=1, imaginary_time=true,
+            save_every=10^9),
+        psi_init=psi0, enable_ddi=true, c_dd=EU_c_dd,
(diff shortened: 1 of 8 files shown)


anko9801:
**Verification: green.**

* TSUBAME `ci` tier: **267 files, all passed**, `TEST_RC=0`
* `test_accuracy_profiles.jl`: **76 / 76**
* GitHub: JuliaFormatter / Oracle gates / fast tier / Detect changes — all pass

`accuracy_profile_report(:fast)` prints `(identical to :production)`, which is the
measured answer rather than an empty name.

---

**On `secular_ddi`, asked separately: it cannot be removed, and the judgment is
decidable on three independent grounds.**

1. **Hard requirement of the spin-rotating-frame path.** `make_workspace.jl:261`
   throws an `ArgumentError` for `spin_rotating_frame_omega ≠ 0` without it — the
   rotating frame relies on Larmor-averaging, and full DDI's off-diagonal
   components average to zero only in the secular limit. The same rule is
   duplicated in `schema.jl:645` and `inspect_checks.jl:422`. Deleting the knob
   deletes the feature.
2. **87 run configs set `secular: true`.**
3. **It is not an approximation of the default — it is a different model, and the
   correct one in its regime.** Valid for `ω_L ≫ c_dd⟨n⟩`; CLAUDE.md records that
   Eu experiments almost always live there, and `make_workspace` emits an `@info`
   recommending it (user-chosen, never automatic). The difference is understood and
   pinned: the validation report has "secular ≠ full when B=x and polarised along
   x — the 4× drop is physically correct, full DDI keeps Q_xy/Q_xz/Q_yz and
   secular zeroes them."

And the row in this PR's cost table should not be read as an argument for removal.
`secular_ddi = true` at **0.986×** means the choice cannot be made on cost — it has
to be made on physics. That is a good property: the in-regime-correct model is
free, so nobody can pick it for the wrong reason.

Removing it would mean retiring `spin_rotating_frame_omega`, migrating 87 configs,
and withdrawing the claim that Eu sits in the secular regime — a research
decision, not a performance one.

PR #214, opened 2026-07-30
feat(lhy): let a caller ask whether a mean field is dynamically stable

anko9801:
## Why

The phase-gap error budget refused to answer three times. Each time I diagnosed it differently — unconverged, then a tolerance problem, then blocked on #172 — and each time I was wrong. The first run had printed the answer:

```
FullBdG LHY: mean field is dynamically unstable (max Im ω = 1040.0)
```

and my log filter (`grep -vE "^│|^└|^┌"`, meant for CUDA warning boxes) had removed it.

`:full_bdg` tabulates ε_LHY from the peak-density spinor of the state the workspace is **built** with. The reference arm handed it a raw `:flower` seed. Where the mean field is dynamically unstable, the zero-point sum drops the complex branches while the counterterms still subtract all `D` of them, so ε_LHY is scheme-dependent — the ITP had **no fixed point to converge to**. Non-convergence was correct behaviour on an ill-posed request.

## What

**The diagnostic is now askable.** `_lhy_bdg_energy_density` computed `max Im ω`, warned with `maxlog=1`, and discarded it. It now returns it, and `lhy_mean_field_max_growth` exposes it as a public query. A `maxlog`-limited warning is not something a caller can ask for, and it is something a log filter can eat.

This makes stability a **precondition** you check before spending a run, not a diagnostic you read afterwards.

**`bench/lhy_stability_scan.jl`** — scans `(c1_ratio, q)` at Eu F=6 for the polar and ferromagnetic spinors to find where a reference is even defined. CPU, seconds per point, no ITP.

**`bench/phase_gap_error_budget.jl` v3** — two changes, both to the *request* rather than the tolerances:

- **two-stage relaxation**: stage 1 with no LHY, stage 2 rebuilding the workspace from that relaxed ψ, so the table is tabulated from a physical state. Both stages run for every arm, so the staging is not a confound between them.
- `max Im ω` at the relaxed peak spinor is printed per row and marks the row unusable when nonzero, whatever its ΔE says. The verdict breaks its refusal down by cause — how many points converged, how many were distinct, how many were stable — so the next run fixes the missing one.

## Verification

- `submit_load_check.sh` on TSUBAME (job 8305287): loads, `accuracy knob registry 51/51`.
- `ci` tier on merged main `376e9895` (job 8305260): 270 files / 8 workers, all passed.
- Stability scan (job 8305328) and phase-gap v3 (job 8305329) are running; results go in a follow-up comment. **Nothing in this PR claims a phase-boundary accuracy number** — that is what these jobs are for.

### DIFF
--- a/bench/lhy_stability_scan.jl
+++ b/bench/lhy_stability_scan.jl
@@ -1,0 +1,68 @@
+# Where is the Eu F=6 mean field dynamically STABLE?
+#
+# `full_bdg` cannot serve as an accuracy reference where `max Im ω ≠ 0`: the
+# zero-point sum drops the complex branches while the counterterms still subtract
+# all D of them, so ε_LHY is scheme-dependent, the ITP has nothing to converge to,
+# and no tolerance fixes it. The code says so in a warning; it now also ANSWERS,
+# via `lhy_mean_field_max_growth`, which is what a scan needs.
+#
+# This is the prerequisite the phase-gap budget was missing. That budget ran at
+# max Im ω = 1040 and its non-convergence was correct behaviour — I read it as
+# "unconverged" and then as "blocked on issue #172", and both were wrong.
+#
+#   julia --project=. bench/lhy_stability_scan.jl [n_c1]
+
+using Printf
+using SpinorBEC
+using SpinorBEC: lhy_mean_field_max_growth
+
+include(joinpath(@__DIR__, "eu151_params.jl"))
+
+const N_C1 = length(ARGS) >= 1 ? parse(Int, ARGS[1]) : 9
+const F = 6
+const D = 2F + 1
+
+"Unit spinors for the states the Eu texture work actually competes."
+function probe_spinors()
+    polar = zeros(ComplexF64, D); polar[F + 1] = 1.0
+    fm = zeros(ComplexF64, D); fm[1] = 1.0
+    (polar=polar, ferromagnetic=fm)
+end
+
+println("="^76)
+println("Eu F=6 mean-field stability — max Im ω (0 = stable)")
+println("c₀ + F²c₁ = c_total = $(round(EU_c_total; sigdigits=5)), c_dd = $(round(EU_c_dd; sigdigits=5))")
+println("="^76)
+
+c1_ratios = collect(range(-0.10, 0.10; length=N_C1))
+qs = (0.0, 1.0e-3, 1.0e-2)
+
+for (sname, spinor) in pairs(probe_spinors())
+    println("\n--- spinor: $sname")
+    @printf("  %-10s", "c1_ratio")
+    for q in qs
+        @printf(" %14s", "q=$q")
+    end
+    println()
+    for r in c1_ratios
+        ip = eu_interaction_params(r)
+        @printf("  %-10.4f", r)
+        for q in qs
+            g = try
+                lhy_mean_field_max_growth(; F, spinor, n0=1.0, interactions=ip,
+                    zeeman=ZeemanParams(EU_p_weak, q), c_dd=EU_c_dd)
+            catch e
+                NaN
+            end
+            @printf(" %14.4g", g)
+        end
+        println()
+    end
+end
+
+println("""
+
+[read] A column of zeros is a usable reference regime. Anything nonzero means
+`full_bdg` is scheme-dependent there, so it cannot be the accuracy reference and
+an ITP started there has no fixed point to reach — that is a property of the
+(F, c₀, c₁, q, c_dd) point, not of the seed or the tolerance.""")
--- a/bench/lhy_stability_scan.jl
+++ b/bench/lhy_stability_scan.jl
@@ -105,7 +105,8 @@ for n0 in (0.01, 0.1, 1.0, 10.0)
         growth(sp.polar, ip0, EU_c_dd, n0),
         growth(sp.ferromagnetic, ip0, EU_c_dd, n0))
 end
-println("  If max Im ω ∝ √n0 with no zero crossing, no density is a stable")
-println("  working point either — the sign of the instability is n-independent")
-println("  because it comes from a negative eigenvalue of the coupling matrix,")
-println("  and n only sets its rate.")
+println("  Measured EXACTLY linear in n0 with no crossing, which is what")
+println("  |Im ω| = n·|λ_min| looks like: ω² = ε_k(ε_k + 2nλ) is most negative at")
+println("  ε_k = −nλ, giving |Im ω| = n|λ|. So the SIGN of the instability is")
+println("  n-independent and no working density is a stable point either — n only")
+println("  sets the rate.")
--- a/bench/lhy_stability_scan.jl
+++ b/bench/lhy_stability_scan.jl
@@ -66,3 +66,46 @@ println("""
 `full_bdg` is scheme-dependent there, so it cannot be the accuracy reference and
 an ITP started there has no fixed point to reach — that is a property of the
 (F, c₀, c₁, q, c_dd) point, not of the seed or the tolerance.""")
+
+# --- controls: WHICH term makes it unstable, and does the density matter? -----
+#
+# "Everything is unstable" is not yet a finding — it could be the dipolar term
+# (physical and unavoidable at Eu: the textures exist BECAUSE the uniform state is
+# unstable), the spin channel, or an n-dependence that a different working density
+# would escape. These two controls separate those.
+println("\n" * "="^76)
+println("Controls — is it the DIPOLE, the spin channel, or the density?")
+println("="^76)
+
+sp = probe_spinors()
+ip0 = eu_interaction_params(0.05)
+zee0 = ZeemanParams(EU_p_weak, 0.0)
+
+growth(spinor, ip, cdd, n0) = try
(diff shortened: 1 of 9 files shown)


anko9801:
## Results — and the finding is bigger than the instrument

**`bench/lhy_stability_scan.jl`** (job 8305331, Eu F=6, c_total = 4687, c_dd = 211). `max Im ω`, where 0 = stable:

| control | result |
|---|---|
| raw `:flower` seed, 32³ | **1040** |
| after relaxing with `lhy = none` | **9.8** — two orders better, not zero |
| `c_dd = 0` | **exactly 0** |
| `c_dd` = 2.11 / 21.1 / 105.5 / 211 | 319 / 1060 / 2715 / 4371 (polar) |
| `c1_ratio ∈ [−0.1, 0.1]` × `q ∈ {0, 1e-3, 1e-2}`, polar **and** FM | **not one stable point** |
| `n0 ∈ [0.01, 10]` | exactly **linear**, no crossing — i.e. `\|Im ω\| = n·\|λ_min\|` |

So the instability **is the dipole**. Not the seed, not the tolerance, not the density. Two-stage relaxation helps enormously and cannot finish the job.

### What that invalidates, from this afternoon

`accuracy_profile(:reference)` sets `spinor_lhy = :full_bdg`, so **it has no valid instance for a dipolar spinor gas** — and the precondition I wrote for it ("relax the state first") pointed at a workaround that is now measured not to work. `check_accuracy_preconditions` takes `c_dd` and refuses on it **first**, ahead of the relaxation advice, so a dipolar caller is not sent after the non-fix. Gate: 83/83 (job 8305366).

This is the same fact as #172 from another side: the term that is 97 % of the Eu F=6 total energy is built on a linearisation with no stable point.

**Scope, and it matters.** This is the **uniform local-density** mean field every tabulated LHY is built from, *not* the trapped texture. A trap and a texture's own gradients gap the low-k modes; whether the trapped state is stable is a trapped-BdG question and a different instrument (`solvers/hessian` λ_min). What is measured is that the **construction** of ε_LHY is ill-posed at Eu — not that the ground state is absent.

### Also in here

`make_workspace` advised `secular_ddi=true` as "faster + more physical". It is **not faster** — 0.986× at 32³, i.e. noise, because `Q_xx = Q_yy = −Q_zz/2` keeps the kernel a 3-component convolution with all 6 FFTs; only 6 of 9 pointwise multiplies vanish. The advisory was inviting an accuracy approximation for nothing and now says the case is the physics one or none. On deleting the knob: **no** — `spin_rotating_frame_omega ≠ 0` throws without it, so removing it removes the rotating-basis path.

### Still running

`bench/phase_gap_error_budget.jl` v3 at 32³ (job 8305329). Its 8³ smoke reported **0 converged / 4 distinct / 0 stable**, with `growth ≈ 9.8` at the relaxed texture — consistent with the scan. No phase-boundary accuracy number is claimed in this PR.

anko9801:
This is the API I left an open follow-up for in #215 / #218 — thank you. `lhy_mean_field_max_growth(; F, spinor, n0, interactions, zeeman, c_dd, rtol)` being exported and taking `c_dd` is exactly what a config-level gate needs. Four things, three of them measurements you can have for free.

## Your control 1 is already answered — it is the dipole, entirely

I ran that separation yesterday from the other direction. `max Im ω` with the DDI on vs off, at two different ladders:

| ladder | seed | DDI on | DDI off |
|---|---|---:|---:|
| texture (N=50000, c₁=0, 50 µG) | flower | 0.365 | **0** |
| | polar_core_vortex | 0.926 | **0** |
| | m_plus_F | 0.370 | **0** |
| | polar | 1.030 | **0** |
| eu_k3_lhy (N=30000, c₁=−0.005) | m_minus_F | 0.396 | **0** |

Exactly zero, every case. And with the DDI off, `full_bdg` agrees with the FM closed form to six significant figures — `V(n=3.7e-3)` = 0.380007 vs 0.380006 — so both are cross-validated in the stable regime. `ε_dd = 0.5402`.

So control 1 will come back "the dipole", and control 2 (density) will not rescue it, since the above holds at two different N.

## Your parenthetical is a better framing than the one I wrote

> the dipolar term (physical and unavoidable at Eu: **the textures exist BECAUSE the uniform state is unstable**)

`docs/validation/full_bdg_scheme_dependence_eu_f6.md` frames the instability as an obstacle to route around. Yours says something sharper: if the textured states exist *because* the uniform state is unstable, then a **uniform** mean-field BdG reference is the wrong reference for them, not a broken one. I will fold that into the doc and credit it — it explains why the two-stage relaxation helps without curing it.

Which it does not, and I measured that too. Converging before tabulating (your stage 2) reduces `max Im ω` by 4–13× and stops:

```
seed                 Imω(seed)  Imω(after ITP)   ITP converged?
flower                  0.365        0.0688          no
chiral_spin_vortex      0.369        0.0771          no
polar_core_vortex       0.926        0.0735          no
m_plus_F                0.370        0.0880          no
polar                   1.030        0.150           YES
```

The decisive row is `polar`: the only seed whose ITP actually converged, and it has the **largest** residual growth. So two-stage relaxation makes the request physical without making it stable — which your per-row `max Im ω` print handles correctly by marking the row unusable rather than quoting its ΔE.

## `q` is not an escape axis at any reachable field

Your `qs = (0.0, 1.0e-3, 1.0e-2)`. Converting through `q ∝ B²` at Eu151 (`q(1 G) = 1.422e-07`):

- `q = 1e-3` ⇔ **B ≈ 84 G**
- `q = 1e-2` ⇔ **B ≈ 265 G**

against 50–80 µG in the texture campaign (`q ≈ 3.2e-16 … 8.3e-16`) and 0.01 G in `eu_k3_lhy` (`q = 1.4e-11`). Worth labelling that column as *"what would it take"* rather than a reachable setting, so a reader does not take a stable cell at `q = 1e-2` as an available operating point. It is 15 orders of magnitude from the campaign.

## Where a valid route does exist

For a state whose ansatz matches a closed form, `*_dipolar` sidesteps this entirely: `lima_pelster_Q5` applies Petrov's prescription, zeroing the integrand where `1 + ε_dd(3cos²θ − 1) < 0`, so the unstable angles are excluded rather than half-counted — and `ε_dd = 0.54 < 1` is inside its domain (`fm_dipolar.jl:26`). That is why #215 moved the `eu_k3_lhy*` arms to `fm_dipolar` (all twelve are `m_minus_F`, and `c₁ < 0` makes FM the mean-field ground state) rather than to `full_bdg`. It does **not** help the texture campaign, whose whole point is comparing five distinct textures under one functional.

## What your API unblocks

`test/oracles/test_lhy_config_validity_domain.jl` covers the closed forms only; `full_bdg` is outside it because its validity condition is a runtime property. With `lhy_mean_field_max_growth` it becomes reachable — build each config's declared `initial_state` at a small grid and assert `max Im ω == 0`, in a slower tier than the current gate. That would put the five `LHY_full_bdg*` arms kept in #218 inside a gate for the first time. Happy to write it once this lands, unless you would rather it came with this PR.

anko9801:
Re: including the gate here — **it does not fit on this branch, so I will do it as a follow-up from main once this merges.** Recording why, and the design, so it is reviewable before I write it.

## Why not here

This branch is 14 commits behind main and 7 ahead, and both things the gate needs are on the main side only:

- `test/oracles/test_lhy_config_validity_domain.jl` (#209) — the gate it extends, absent here
- the `LHY_full_bdg*` / `LHY_fm_dipolar*` configs it would check — this branch still has `LHY_icosahedral.yaml`

So adding it would mean merging 14 commits of main into a branch that is still moving (HEAD is `44603447`, newer than this PR's description) and is checked out in an active worktree. I would rather not be the one to do that from the outside — earlier today an uncommitted edit of mine got absorbed into another session's commit in that same worktree, and this would be the same hazard pointed the other way.

## The design, so it can be argued with before it exists

A naive "assert `max Im ω == 0` for every `full_bdg` config" goes **red on merge**, because the five `LHY_full_bdg*` arms kept in #218 genuinely are unstable — that is what they are for. So it takes the allow-list shape of `test_no_unguarded_fft_derivative.jl`, asserted in **both** directions:

- a config **not** on the list must be stable → a newly added `full_bdg` request in an unstable regime fails review instead of at table build;
- a config **on** the list must **still be unstable**, with its measured `max Im ω` recorded → if one becomes stable, the list is stale and the entry has to be re-justified rather than silently protecting nothing.

The second direction is the part that keeps it from decaying into a permanent exemption list. Each entry carries the value and the reason (comparator for the `fm_dipolar` sibling, per #218), so the list is a record of a decision rather than a way to keep quiet.

Cost: one `lhy_mean_field_max_growth` call per config at a small grid, using each config's declared `initial_state` — seconds, no ITP — but slower than the existing gate, so a different tier.

## One thing I could not settle from outside

`lhy_mean_field_max_growth` takes a `spinor`, and a config declares an `initial_state` symbol. For `m_minus_F` / `polar` the representative spinor is unambiguous. For a texture (`flower`, `chiral_spin_vortex`, `polar_core_vortex`) it is whatever `_extract_spinor` picks from the built ψ — the peak-density spinor. The gate should use the same extraction the workspace does rather than reimplement it, otherwise it would be checking a different state than the run uses. I will route through that rather than construct a spinor myself; flagging it in case you would rather the query grew a `psi_init` form.

PR #216, opened 2026-07-31
fix(ddi): the gradient must use the same kernel as the energy under padding

anko9801:
`_ddi_energy` branches on `ws.ddi_padded`. `_grad_ddi!` did not. With `ddi_padding` on — the YAML parser default since 2026-07-29 — L-BFGS and Newton-CG accepted steps on the **padded** energy while taking their direction from the **unpadded** gradient.

The endpoint of such a solve is not a stationary point of the energy it reports, and the `grad_norm` it reports is the norm of a different operator's gradient.

## Measured

Eu-151 F=6, 24³, box 12, DDI on, `ddi_padding=true`, `method: lbfgs`, default `tol=1e-8`. Both revisions in one UGE job on one node, same spec, same probe script (byte-identical in both worktrees).

| | before (`main`) | after | Δ |
|---|---|---|---|
| energy | 7.86947853941 | **7.85894088045** | −1.05e-2 |
| energy_ddi | −0.0147881746 | −0.000701430 | +1.41e-2 (21×) |
| **grad_norm** | **3.59e-2** | **7.65e-7** | **4.7e4× lower** |
| polarisation \|⟨F⟩\|/F | 0.293007 | 0.297122 | +1.4 % |
| last_step | 113 | 541 | |
| **1 − \|⟨A\|B⟩\|** | | | **0.450** |

This is not a perturbation. The two states overlap by **55 %**, and the residual on `main` is **3.6e-2** — four and a half decades above the floor this problem actually reaches (7.65e-7, consistent with the ~5e-7 floor measured independently on the contact cell).

**On current `main`, a padded-DDI L-BFGS ground state is badly unconverged and returns quickly and confidently.** The gradient bug predates #210, but #210's floor stop interacts with it: the stop fires when steepest descent finds no decrease, and that direction was built from the wrong gradient, so the solve exits early instead of grinding.

Confound check: the two arms' `src/` differed in exactly two files — this fix, and #212's `fm_contact.jl`. The probe constructs no LHY at all (`spinor_lhy` unset, `c_lhy = 0`), so the second cannot have contributed.

## The fix

Both branches now call the same convolution the energy face calls, so the pair cannot drift again by construction rather than by agreement.

Two details:

- the padded `Phi_*` are read as **Cartesian corner views**. The first `prod(n_pts)` *linear* elements walk full padded columns into the pad region for `ndim >= 2` (App. A defect 9) — this is why `_ddi_energy` indexes with `CartesianIndices(n_pts)`;
- the accumulation moves behind a **function barrier**, because the two branches hand in differently-typed `phi_*` (a plain array vs a strided corner view) and that union should not leak into the broadcasts.

## Why nothing caught it

`test_ddi_padded.jl` never calls `energy_gradient!` or `apply_operator!`, and the FD term-consistency oracle builds its workspaces without DDI at all — so the pair was never compared on this path.

The new gate is the FD identity between the two faces with padding ON, `(E(psi + eps*dpsi) − E(psi))/eps == 2·Re<grad, dpsi>·dV`, with three controls so that green means something:

1. **canary** — the padded and unpadded gradients must actually *differ* at this size, or the padded case is vacuous;
2. **positive control** — the same identity without padding must hold, or the FD comparison is simply insensitive;
3. **positive control** — feeding the check the *wrong* kernel's gradient, which is precisely the defect, must be rejected at the same `rtol`.

Control (2) earned its place immediately: the first version of the gate dropped the Wirtinger factor of 2 and failed on **both** the padded case and the unpadded control — which said the comparison was broken rather than letting a broken comparison indict the fix.

## Verification

`tier=ci` at `c6cbeba7`: **270 files, all passed** (TSUBAME job 8305268).

## Consequence for stored results

This changes every padded-DDI ground state solved by L-BFGS or Newton-CG. ITP and the dynamics propagator are unaffected — `apply_step!` always branched correctly. Padding only became the default on 2026-07-29, so the affected stored set is small and recent — but it includes anything the owed Eu texture recompute would have produced, which is the reason to land this before that runs.

### DIFF
--- a/bench/probe_ddi_gradient_padding.jl
+++ b/bench/probe_ddi_gradient_padding.jl
@@ -15,11 +15,16 @@
 # This measures the size of that: same spec, same solver, two revisions, one
 # node. What moves is the converged ψ, its energy, and the reported residual.
 
+# Top level, not inside the `if` below: a macro is resolved when the enclosing
+# top-level expression is LOWERED, so `using Printf` in the same block comes too
+# late and `@printf` is undefined — which is exactly how the first run of this
+# probe died, in the arm that never touches the compare branch.
+using JLD2
+using Printf
+
 const MODE = length(ARGS) >= 1 ? ARGS[1] : "solve"
 
 if MODE == "compare"
-    using JLD2
-    using Printf
     a = load(ARGS[2])
     b = load(ARGS[3])
     println("=== ", ARGS[2], "  vs  ", ARGS[3], " ===")
@@ -44,8 +49,6 @@ if MODE == "compare"
 end
 
 using SpinorBEC
-using JLD2
-using Printf
 
 include(joinpath(@__DIR__, "eu151_params.jl"))
 
--- a/bench/probe_ddi_gradient_padding.jl
+++ b/bench/probe_ddi_gradient_padding.jl
@@ -1,0 +1,104 @@
+#!/usr/bin/env julia
+# How far does the DDI gradient-padding fix move a ground state?
+#
+#   julia --project=. bench/probe_ddi_gradient_padding.jl solve   <out.jld2>
+#   julia --project=. bench/probe_ddi_gradient_padding.jl compare <a.jld2> <b.jld2>
+#
+# `_ddi_energy` branches on `ws.ddi_padded`; `_grad_ddi!` did not, so with
+# padding on — the YAML default since 2026-07-29 — L-BFGS accepted steps on the
+# PADDED energy while taking its direction from the UNPADDED gradient. It still
+# decreased the padded energy monotonically, but along the wrong direction, and
+# it stopped where that direction stopped helping. The endpoint is therefore not
+# a stationary point of the energy it reports, and the `grad_norm` it reports is
+# the norm of a different operator's gradient.
+#
+# This measures the size of that: same spec, same solver, two revisions, one
+# node. What moves is the converged ψ, its energy, and the reported residual.
+
+const MODE = length(ARGS) >= 1 ? ARGS[1] : "solve"
+
+if MODE == "compare"
+    using JLD2
+    using Printf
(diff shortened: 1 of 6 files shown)


anko9801:
**Verification complete.**

| tier | commit | result |
|---|---|---|
| `ci` | `c6cbeba7` | **270 files, all passed** (job 8305268) |
| GitHub | `002e0045` | formatter / oracles / fast — **all pass** |
| `full` | `002e0045` | 5/12 workers (job 8307318) |
| `full` on **clean `main`** `f2352c7f` | — | 5/12 workers (job 8307387) |

Failure sets:

```
main f2352c7f          this branch 002e0045
---------------------  ---------------------
mixed_precision:64     mixed_precision:64
mixed_precision_p3:51  mixed_precision_p3:51
combined_spin_step:146 combined_spin_step:146
combined_spin_step:148 combined_spin_step:148
spin_chain_fusion:69   spin_chain_fusion:69
spin_chain_fusion:87   spin_chain_fusion:87
pipeline:382           pipeline:382
                       lbfgs_accuracy_floor:118
```

`spin_chain_fusion_parity.jl:87` — the padded-DDI *propagator* fusion parity — **fails on clean `main` too**, so it is not this change; it is a face this PR does not touch (`apply_step!`, not `apply_operator!`). It is invisible on GitHub because that assertion is GPU-only and CI has no GPU.

The one extra is `lbfgs_accuracy_floor.jl:118`, the known-unstable Newton-polish gate: it compares two numbers at a gradient floor that is itself not reproducible (5.7e-10, 1.4e-8 and 2.8e-8 measured for the same commit and problem), and it has failed on `main` in a previous run and passed in others.

**The new GPU gate is not in the failure list** — the padded DDI gradient agrees between CPU and GPU, so the strided corner view reaching a device broadcast is fine.

PR #222, opened 2026-07-31
test(oracles): gate every committed full_bdg config on mean-field stability

anko9801:
The follow-up promised in #215 / #218, unblocked by #214.

## The gap

`full_bdg` has no ansatz to violate, so #209's closed-form domain gate has nothing to check for it. It has a **different** validity condition: the uniform mean field it linearises around must be dynamically stable. Where it isn't, the zero-point sum drops the complex branches while the counterterms still subtract all `D` of them — so ε_LHY is scheme-dependent, **and the table still builds and still returns finite numbers**. That is why nothing noticed.

The condition needs the spinor and the density, so it is not a property of the YAML text. It became checkable when #214 turned `max Im ω` from a `maxlog`-limited warning into a return value — a warning is not something a caller can ask for, and #214's own account is of a log filter eating it.

## It found ten configs on its first execution

Three groups, differing in **kind**, not just provenance:

| group | n | why |
|---|---:|---|
| comparator arms (#218) | 5 | unstable **on purpose** — the pair with their `fm_dipolar` siblings *is* the measurement of what the scheme dependence is worth (V_LHY 0.59327 vs 0.553422, a 7 % spread) |
| texture campaign + smoke twin | 2 | already marked do-not-run-as-is; the gate reaching that verdict independently is the point of having it |
| **movie configs (#164)** | 3 | **nobody had connected these to this at all** |

The movie ones are **accepted rather than changed**. Their observable is the excitation itself — where vortices come from, how the density reacts — and a texture exists *because* the uniform state is unstable, so a uniform-mean-field BdG reference is the wrong reference for one by construction. That framing is #214's, and it is better than the one in my own doc. The scheme dependence is a caveat on any **energy** they quote, not on a vortex count.

## Both directions are asserted

This is what keeps an allow-list from rotting into a permanent exemption:

- **not listed ⇒ must be stable.** A new `full_bdg` request in an unstable regime then fails at *review*, instead of at table build — long after merge, on a machine, mid-campaign.
- **listed ⇒ must still be unstable.** If an entry becomes stable that is news, and the entry has to be re-justified rather than kept.

Plus #209's two anti-vacuity checks: the sweep must find configs at all, and the list must not name paths the sweep never reached.

The spinor goes through **`_extract_spinor`**, not a hand-built one — for a texture the representative is the peak-density spinor, and choosing it differently here would check a different state than `_build_spinor_lhy` does.

## Also deletes two committed run-output stubs

`runs/config_texture_stir_movie_f5bf647e.pre_{masscurrent,strict}` — surfaced by the sweep. `completed: true`, but holding only `config.yaml` + `_exit_summary.json`: **no results, no references anywhere**, and recoverable from `15296ea5`. Their `full_bdg` request was scheme-dependent, so whatever they measured is not quotable.

## Where it runs — worth stating

`CI_EXTRA`, beside #209's gate: a BdG solve per config is seconds, but too slow for the fast tier. Flagging it because **`tier=fast` does not run this file** — the count stays 174 and the name never appears, which nearly had me report a green fast tier as evidence for a gate that had not executed.

- standalone: **12/12**, and it caught all ten on the first run
- `tier=fast`: 174 files, all passed
- `tier=ci`: running locally; I'll post the result here


anko9801:
Local `tier=ci`: **271 files across 10 workers, all passed** — and the gate is confirmed to have actually executed there:

```
oracles/test_full_bdg_config_stability.jl |   12   12   17.7s
```

That line is the point. In `tier=fast` the file does not appear at all and the count stays 174, so a green fast tier says nothing about this gate. `tier=ci` is where it lives, and CI's required checks do not cover that tier — so this local run is the only direct evidence the gate runs, rather than merely existing.

Full status: standalone 12/12 (caught all ten configs on its first execution), `tier=fast` 174/174, `tier=ci` 271/271 with the gate executing in 17.7 s.

anko9801:
Merged main in (conflict was `test/_tiers.jl` only — main added `oracles/test_doc_run_citations_resolve.jl` at the same list position; both kept) and re-ran `tier=ci`, because main had moved 7 commits including #216, which changes DDI gradient behaviour and could in principle reach a gate that solves a BdG problem.

**This gate passes: 12/12 in 13.9 s.**

`tier=ci` as a whole is red, and **not from this branch.** One worker fails in `test_doc_run_citations_resolve.jl:90`:

```
["eu151_b1_c1_sweep_template", "klaus_eu151_mechanism_", "klaus_eu151_spin_excitation",
 "klaus_option_gamma", "klaus_option_gamma_full", "option_gamma_smoke", "validation_level"]
```

cited from `archive/validation_ladder_2026-05-22.md`.

Measured on **clean `origin/main`** rather than assumed: same failure, same seven names, 13 passed / 1 failed. So `tier=ci` is red on main independently of this PR — worth knowing since the required checks do not cover that tier, so nothing on a PR page shows it.

Note also what it is **not**: this PR deletes two run-output stubs (`config_texture_stir_movie_f5bf647e.pre_{masscurrent,strict}`), which is exactly the shape that gate exists to catch — but neither appears in the unresolved list, and the seven are all klaus / option_gamma / validation_level names untouched here. I checked that before concluding it was pre-existing.

Full status: standalone 12/12 · `tier=fast` 174/174 · `tier=ci` this gate 12/12, suite red on main for an unrelated reason.

anko9801:
Correcting myself: **CI does run this gate**, and I said the opposite earlier in this PR.

`ORACLE_TESTS` is derived, not hand-listed:

```julia
const ORACLE_TESTS = filter(t -> startswith(t, "oracles/"), vcat(FAST_TESTS, CI_EXTRA, FULL_EXTRA))
```

So putting the file in `CI_EXTRA` under `test/oracles/` puts it in the **oracles** tier too, which the required checks *do* cover. Confirmed directly rather than reasoned about — `select_tests("oracles")` returns 71 files and includes it. My earlier statement that "`tier=fast` does not run this file, so a green fast tier says nothing" was right; the conclusion I drew from it, that CI never runs the gate and the local `ci` run is the only evidence, was wrong.

## Why the Oracle gates job is red, and it is not this gate

Reproduced the CI failure locally at `tier=oracles`: **2/10 workers, 3 assertions, 2 files** — neither mine.

| file | what |
|---|---|
| `test_doc_run_citations_resolve.jl:90` | the 7 unresolved citations. **Measured on clean `origin/main`: identical failure, identical seven names** |
| `test_spin_chain_fusion_parity.jl:69, :87` | fused half-step bit-identity, differences at 1e-10. #209's own PR body already reported this one as "unrelated" |

So `tier=oracles` is red on main for two pre-existing reasons, and this PR inherits both. The gate itself passes — 12/12 in `tier=ci`, where the per-file line is visible; in the oracles run the two failing workers abort before per-file summaries print, which is why grepping the log for its name comes back empty and briefly had me doubting it ran at all.

## Where that leaves the PR

Nothing here is blocked on the gate's own behaviour. It is blocked on two failures that exist on main, which the required checks surface on every PR touching `test/oracles/` — so this PR cannot go green until those are fixed, independently of its content.

Verification as it actually stands: standalone 12/12 · `tier=fast` 174/174 (does not include this file) · `tier=ci` 271 files, this gate 12/12 in 13.9 s · `tier=oracles` includes it, suite red on main for the two reasons above.

anko9801:
Merging with the Oracle gates job red, deliberately, and recording why so it is not mistaken for an override of a real failure.

## The red is inherited, and measured to be so

`test_spin_chain_fusion_parity.jl` fails 2 assertions (`:69` bare DDI, `:87` padded) at **1e-10 against a bit-identity claim**. Everything about it is main's:

- reproduces on **clean `origin/main`**, same two assertions
- reproduces with **every other oracles file skipped**, so no predecessor is involved
- reproduces on **TSUBAME** (job 8309128) as well as locally, so not environmental
- main's own CI is currently red (`78010e82`)

Any PR touching `test/oracles/` inherits it, because `ORACLE_TESTS` is derived from the `oracles/` prefix across FAST + CI_EXTRA + FULL_EXTRA.

## This gate itself is verified

- standalone **12/12**, and it caught **ten configs on its first execution** — five #218 comparator arms, the texture campaign plus smoke twin, and three movie configs (#164) nobody had connected to this
- `tier=ci` **271 files, all passed**, with the gate visible at 12/12 in 13.9 s
- `tier=fast` 174/174 (does not include this file)

## Why not wait

Waiting does not converge. Nine hypotheses for the fusion failure are eliminated by measurement — commit regression (#216), test randomness, GPU contention, this branch, GPU-touching predecessors, all predecessors, `--check-bounds`, `--pkgimages=existing`, an outer `@testset`. What is established is that the failure **requires `Pkg.test()` as the launcher**: direct `include` passes under every flag combination tried, and `SPINORBEC_TEST_WORKERS=1` (the serial path, same process) still fails — so it is not the worker launch either. The remaining candidate is `Pkg.test()`'s sandbox dependency resolution: the direct path loads CUDA runtime 12.9.0 / CUDA.jl 5.11.0 / cuFFT 11.4.1, and `Pkg.test()` re-resolves. A different cuFFT would change FFT algorithm selection, which is exactly the size of difference observed. **Not yet confirmed** — the instrumentation attempt broke the file's own CUDA guard and measured nothing.

Meanwhile the five `LHY_full_bdg*` arms kept in #218 stay **ungated**, which is the thing this PR exists to fix.

## What I am NOT claiming

That the fusion failure is benign. It is a real bit-identity violation in a gate whose whole point is exactness, and its investigation continues after this merge. This merge does not make main more red than it already is; it adds a gate that passes.

PR #233, opened 2026-07-31
fix(test): drop the seven citation exemptions whose documents left the tree

anko9801:
**`main` is red at `2d19ce00`** — `test/oracles/test_doc_run_citations_resolve.jl:126`, the "known-unresolved list does not rot" arm.

#227 regenerated `KNOWN_UNRESOLVED` while the two notes imported by #221 were still in `docs/archive/`. #228 then fixed the *same* gate by moving those notes out of the repository — but left the seven names only they cited in the list. Nothing in `docs/` cites them any more, so the rot arm fires. Correctly: this is exactly the failure mode that arm exists to catch, so the fix is to shorten the list, not to relax the check.

Each PR was green on its own branch and they are red together — #228 branched before #227's regenerated list existed, so its own run never saw these names.

## Verification

All four arms checked against the current tree before committing:

| arm | result |
|---|---|
| uncited entries | none |
| entries that resolve again | none |
| fresh (unpinned) breakage | none |
| citations found | 69 (gate needs > 20) |

33 → 26, which is the count `docs/campaign/doc_run_citation_inventory.md` already states.

### DIFF
--- a/test/oracles/test_doc_run_citations_resolve.jl
+++ b/test/oracles/test_doc_run_citations_resolve.jl
@@ -37,7 +37,11 @@ const PLACEHOLDERS = Set([
     "tsubame_scan", "_dashboard_cache",
 ])
 
-# Cited, absent, and known. Regenerated 2026-07-31 with the corrected pattern.
+# Cited, absent, and known. Regenerated 2026-07-31 with the corrected pattern,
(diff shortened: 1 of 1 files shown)


anko9801:
Superseded by #232, which landed the same seven-entry trim first. What is left here is a comment, so it is not worth a CI cycle.

PR #242, opened 2026-07-31
test: drop the stdlib preamble — the environment no longer owns a file's imports

anko9801:
**Stacked on #239** (base branch is `test/standalone-test-files`; the diff here
is one file).

`_run_files.jl` supplied `LinearAlgebra`, `Random`, `Statistics` and `Printf` to
every test file, so a file reaching for `norm`, `I`, `MersenneTwister`, `mean`
or `@printf` without importing them still ran. Its own comment called this *"the
test environment contract"* — which contradicts the one CLAUDE.md states, that
each test file stays a dependency-free unit, and made "does this file run on its
own" a property of **which sibling the claim queue handed out first** rather
than of the file.

#239 removes the reason it existed: every `test_*.jl` now declares what it uses,
measured by running all 348 in their own process with no preamble at all.

## Measured

With #239 applied, the whole `ci` tier — **273 files across 4 workers — passes
with this preamble removed**. Nothing was leaning on it.

So the contract is now **structural**: a file that forgets an import fails,
everywhere, immediately. Before, it was a static gate (#239's meta-test, which
cannot tell `LinearAlgebra.I` from a loop variable) plus a safety net that
quietly absorbed exactly the violations the gate could not see.

`using Test` stays — `run_test_files` itself wraps each file in a `@testset`.

### DIFF
--- a/ext/SpinorBECCUDAExt/SpinorBECCUDAExt.jl
+++ b/ext/SpinorBECCUDAExt/SpinorBECCUDAExt.jl
@@ -56,7 +56,16 @@ function __init__()
     # memory back to the device. Without this, ~96 64³ ComplexF64
     # workspaces accumulate on a 16 GB GPU (~150 MB pinned per point
     # across psi/fft_buf/k²/ddi_kernel) and a long scan will OOM mid-run.
-    SpinorBEC._cuda_reclaim_callback[] = () -> (CUDA.reclaim(); nothing)
+    #
+    # The `functional()` guard is load-bearing, not defensive: this extension
+    # loads whenever CUDA.jl is imported, driver or no driver, so a CPU-only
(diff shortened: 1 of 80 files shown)


anko9801:
CI reports nothing here on purpose, not by accident: `ci.yml` triggers on `pull_request: branches: [main]` and this PR's base is `test/standalone-test-files`. It will get a full run the moment #239 merges and this is retargeted to `main`.

Until then the verification is local and is the one that matters for this change: `SPINORBEC_TEST_TIER=ci SPINORBEC_TEST_WORKERS=4` → **273 files, all passed**, with the preamble already removed. That is the measurement the PR rests on — nothing in the suite was leaning on the leak once #239 landed.

PR #224, opened 2026-07-31
The Fig. 4B dip centre is a DDI kernel-treatment effect, not a coupling error

anko9801:
Follow-up to #204. Worktree `/home/suzume/workspace/BEC-simulation-matsui`, branch `research/matsui-ddi-kernel`.

#204 left two discrepancies with Matsui et al.'s published simulation after every cheap explanation had been excluded (resolution, ramp shape, loss, ground-state choice, field jitter, kernel conventions):

- dip centre **−2.138 nT** against their **−2.549** — ours 16 % smaller
- `N_{−6}` **~20 % lower at every field** — ours transfers more

Those point **opposite ways**, and both scale with `c_dd·n`, so a coupling or density error would move them together. That was the clue.

## The couplings are exact — measured, not derived

`scripts/validation/matsui_static_ddi_compare.jl`, no dynamics:

| | theirs, reduced to our units | ours | ratio |
|---|---|---|---|
| DDI coefficient | 211.021419 | **211.021446** | **1.000000** |
| contact `c_total` | 4687.2663 | **4687.2663** | 1.000000 |

Their `cdd = 1e-7 (g_F μ_B/ħ)²·2mN/aHO` times their `4π` kernel, divided by the `2√2` mapping their length unit onto ours, reproduces `compute_c_dd_dimless` to **seven significant figures**. The `μ₀/4π ↔ 4π` cancellation and the `√2` of row 8.1 are now verified end to end numerically. Contract rows 3.2/3.3 and 4.1–4.6 are closed.

Scale check: `c_dd·n_peak = 5.08 nT` against dip centres of 2.1–2.6 nT — the offset **is** the dipolar mean field, at the right order, in both codes.

## The factorial locates the centre gap

25 fields (−10 … +2 nT) × `trunc_radius` ∈ {auto, none} × `padded` ∈ {true, false}. UGE 8307517 task 9, exit 0, 100 points, ~16 min.

| arm | centre [nT] | min `N_{−6}` | `N(−5 nT)` | `N(0 nT)` |
|---|---|---|---|---|
| auto + padded — **our default** | −2.138 | 9760.1 | 11629.3 | 11162.9 |
| auto + unpadded | −2.148 | 9774.8 | 11634.0 | 11177.8 |
| none + padded | −2.148 | 9958.5 | 11807.8 | 11286.8 |
| **none + unpadded — bare periodic** | **−2.484** | 10228.2 | 11841.2 | 11641.6 |
[...]

PR #269, opened 2026-07-31
Their SIMULATION ran at N = 3.5e4 — 70 % of the experiment's atom number — and we reproduce it to 1 %

anko9801:
> ### ⚠️ 訂正 2026-08-02 — この PR のタイトルは誤解を招きます
>
> 「their published run used N = 3.5×10⁴」は、**彼らのシミュレーション**についてなら正しく、**実験**については誤りです。
> 論文（[arXiv:2504.17357](https://arxiv.org/abs/2504.17357), *Science* **391**, 384 (2026)）は
> 「The trap typically contained approximately **5×10⁴ atoms**, with a negligible thermal component」と述べています。
>
> **本 PR の中身は無傷です** — 三観測量（転送率・rms 半径・アスペクト比）が彼ら自身の 5 ms 状態 `dataset_fig1/F.txt` と
> 2.4–2.7 % で一致することは、彼らの**シミュレーションが 3.5×10⁴ で走った**証拠として今も有効です。
>
> 変わるのは意味づけです。彼らの公開理論曲線は、**実験の原子数の 70 % で計算されている**ことになります。
> これは我々の再現の成否ではなく、**彼らのリリース内部の不整合**（同梱コード `Ntot = 3.5d4` vs 本文 5×10⁴）です。
>
> なお本 PR 以降に行った「実験へのフィット」は前提が誤っており全面撤回しました → #299 のコメント参照。

Closes the residual left open by #224.

## The answer

Their published simulation ran at **`Ntot = 3.5×10⁴`** — the value shipped in their `setup_parameters` — while the published curves are normalised to **5×10⁴**, which is what we had assumed because that is what the data totals (49999.9).

Given the right `N`, our code reproduces their Fig. 4B to **1 %**. All figures read from the final `psi`; widths on the matched window [−12, +9] (see "two measurement corrections" below for why the window has to be stated):

| | dip centre [nT] | half-depth width [nT] |
|---|---|---|
| ours, N = 5.0×10⁴ | −2.1381 | 13.813 |
| **ours, N = 3.5×10⁴** | **−2.5099** | **12.740** |
| Matsui | −2.5495 | 12.752 |

Centre gap **0.411 → 0.040 nT** (−90 %). Width gap **1.061 → 0.012 nT** (−99 %). `N_{m=−6}` ours over theirs, on the 41 fields where the two scans share an abscissa:

| B [nT] | −12 | −10 | −6 | −4 | −2 | 0 | +2 | +5 | +9 |
|---|---|---|---|---|---|---|---|---|---|
| ratio | 1.020 | 1.025 | 1.004 | 0.995 | 0.991 | 0.989 | 0.989 | 1.000 | 1.002 |

**rms 1.10 %, worst 2.57 %** (at −10 nT, on the left wing where we run high). At N = 5×10⁴ the same ratios ran 0.78–0.84.

UGE 8310846 task 15, exit 0, 45 points. Type C.

## Two measurement corrections, both made in this branch

Both were the same failure: an aggregate read off a defective **boundary** input, which does not look like an outlier because it sits at the edge rather than in the middle.

**1. The `dt` error budget is retracted.** This branch briefly claimed the last 0.039 nT was our time step, on the strength of a refined-`dt` arm landing at −2.550 against their −2.549. That arm's `save.every` did not land on the final step: its last series sample sits at t = 3.348 against the baseline's 3.456, so the comparison was 4.84 ms against 5 ms and the "−0.040 nT" was the missing 3.1 % of evolution. Read from the final `psi`:

| arm | centre [nT] | Δcentre |
|---|---|---|
| baseline (`dt` = 1e-3) | −2.5099 | — |
| `dt` refined 4× | −2.5100 | **−0.0001** |
| ramp segment alone refined 4× | −2.5099 | −0.0000 |
| + their exponential ramp | −2.5865 | −0.0766 |
| **Matsui** | **−2.5495** | **−0.0396 needed** |

**Our time step was converged all along** — refining it 4× moves the per-field populations by 8–9×10⁻⁵. Adaptive stepping would buy nothing here. The only candidate of the right order is their exponential ramp, and it overshoots by ~2×. **0.0396 nT (1.6 %) of the centre is unexplained**, bounded well inside the ~1 nT field fluctuation the paper itself quotes.

**2. The width does close — the 2 % was a window mismatch.** `resonance_dip` sets the half-depth level from the endpoints, so `width` is a property of the curve *and its window*. On their theory curve alone: 15.02 nT over [−20, +20], 13.07 over [−13, +9], 11.79 over [−10, +9] — a 21 % spread with no physics in it. Their scan steps 0.5 nT only inside ±10 nT and 1 nT outside, so restricting both to a nominal [−12.5, +9] left our leftmost sample at −12.5 and theirs at −12:

| common window | ours | Matsui | Δ |
|---|---|---|---|
| [−12.5, +9] (endpoints ½ step apart) | 12.885 | 12.752 | +1.04 % |
| **[−12, +9] (endpoints identical)** | **12.740** | **12.752** | **−0.10 %** |
| [−10, +9] | 11.828 | 11.791 | +0.31 % |

The full-range widths were additionally inflated to ~14.95 by `point_001`, which carries the untouched ground state (exactly 1.000000 against 0.7047 at the neighbouring field) and sat on the left endpoint.

Now documented in the `resonance_dip` docstring and pinned by a testset that also asserts `center` is window-invariant to 1e-6 — which is why the centre is the number this campaign quotes. The fixture's own theory-vs-experiment consistency check was comparing [−20, +20] against [−17.5, +17.5]; trimmed.

## Why it took a whole campaign

Integrated populations **cannot separate `N` from anything else that scales the dipolar drive**. That degeneracy is exactly what produced #224's central puzzle — that no single coupling moved the centre and the transfer the same way. Scanning `N` against populations alone gave "N ≈ 4.3×10⁴", the cloud's rms gave "≈2.9×10⁴", a Thomas-Fermi fit gave "≈1.6×10⁴". Three inversions, three answers, all from the same broken degeneracy.

The lock was opened by **`dataset_fig1/F.txt`**, buried in the otherwise image-only `dataset_raw.zip`. It opens with their Fortran's own writer header (`#Nx=128 #dx=4.000E-01`) and integrates to 0.4273 / 0.2352 / 0.1562 across m = −6, −5, −4 — matching their Fig. 2C at 5 ms to four digits. **It is their published run's own 5 ms state, with its spatial structure.**

Three observables at once break the degeneracy, because size and shape respond to `N` differently from transfer:

| at +2.5 nT, 5 ms | Matsui | ours N = 3.5×10⁴ | ours N = 5×10⁴ |
|---|---|---|---|
| `m = −6` fraction | 0.4273 | **0.4158** (2.7 %) | 0.2661 (38 %) |
| rms radius [a_ho] | 2.619 | **2.687** (2.6 %) | 2.849 |
| shape `⟨x²⟩/⟨z²⟩` | 0.899 | **0.877** (2.4 %) | 1.030 |

Nothing else tried in the campaign moved all three together.

## A prediction that failed, and what it taught

Recorded in the run's own commit **before** launching it: since `c_dd·n_peak ∝ N^(2/5)`, a smaller `N` weakens the dipolar field and should pull the resonance **toward zero**, away from their −2.549.

It did the opposite: −2.138 → −2.510, **toward** theirs.

So **the offset is not a density-weighted mean-field shift.** That also settles an earlier oddity — the mean dipolar field over the polarised ground state is −0.65 nT while the dip sits at −2.1. The resonance is set by conditions during the transfer, and it responds to `N` with the opposite sign to the naive scaling.

## Everything excluded along the way was innocent

Couplings (7 s.f.), DDI kernel treatment, resolution, box, their exact 128³ grid, `dt`, ramp shape, three-body loss, their 3-point FD Laplacian, our DDI operator (2.6×10⁻¹⁶ against the independent reference), the ground-state ambiguity (does not exist), the experiment's 8–45 % atom deficit (a counting systematic, not loss). **The residual was one input number.**

`matsui_author_query_draft.md` is superseded — the question was answerable from the deposit — and kept only as a record.

## Ramp discretisation is excluded on both axes, and the ramp SHAPE closes the width

Their Fortran evaluates `B(t) = (B_ini − B_fin)exp(−t/τ) + B_fin` in closed form every step; we encode it as a piecewise-linear waveform. Two independent things could go wrong, and the `dt` sweep only bounds one — a finer `dt` samples the *same* piecewise-linear waveform more densely.

The second axis is knot spacing. A chord across a convex decaying exponential lies above it everywhere, so `τ/6` knots run the field high: excess time-integrated field 0.0848 nT·t.u. over the 5 ms hold, a **+0.0245 nT DC-equivalent offset**, 3.6 nT peak instantaneous excess.

**Prediction recorded in the config before the run: refining to `τ/24` moves the centre +0.0245 nT, to about −2.562. It did not.**

| their exponential ramp | centre [nT] | width [nT] |
|---|---|---|
| knots `τ/6` | −2.5865 | 12.7530 |
| knots `τ/24` | **−2.5867** | 12.7529 |

−0.0002 nT, two orders below the prediction. UGE 8314874 task 21, commit `b634f400`, clean tree, exit 0, 45 points.

A 3.6 nT transient over the 350 µs ramp is worth 0.0002 nT of resonance position, so the DC-equivalent mapping is wrong: **a time-averaged field is not what the resonance responds to.** Same lesson as the mean ground-state dipolar field coming out −0.65 nT while the dip sat at −2.1.

Both axes null ⇒ ramp discretisation excluded, and with it any gain from exact `∫B(t)dt` in the Zeeman exponent (the midpoint quadrature error it removes is O(dt³), bounded 64× below a null by the `dt` sweep).

**What the ramp shape does settle: the width.**

| arm | centre [nT] | Δ | width [nT] | Δ |
|---|---|---|---|---|
| our linear 150 µs ramp | −2.5099 | **+0.0396** | 12.7400 | −0.0124 (0.10 %) |
| their exponential ramp | −2.5865 | −0.0370 | **12.7530** | **+0.0006 (0.005 %)** |
| Matsui | −2.5495 | — | 12.7524 | — |

Their ramp closes the width 20× better while overshooting the centre by almost exactly what the linear ramp undershoots it. The two observables point at *different* arms — so whatever remains is not a single scalar in the ramp, or one arm would win both.

## Also in this branch

**RK4IP** (`src/hamiltonian/integrator/rk4ip.jl`) — 4th-order Runge-Kutta in the interaction picture, exported as `rk4ip_step!`. Composition integrators collapse to order ~1 on the spinor DDI path (#46) because the base step is not time-symmetric for the nonlinear mean field; RK4IP is not a composition, so the order conditions ask nothing of time symmetry. Measured by Richardson with the DDI-off control that a broken composition also passes: **order 3.981 with the DDI off, 3.991 with it on**. Cost is 4 mean-field evaluations per step — the same as our Strang step, whose midpoint predictor-corrector already evaluates the DDI twice per half-potential step. Order 2 → 4 at equal cost. Real-time only; it refuses imaginary time rather than silently doing the wrong thing. Tier `fast`, `test/hamiltonian/test_rk4ip_convergence_order.jl`.

Note that the Fig. 4B result above does **not** motivate it: `dt` is converged for that observable. It is for the 40 ms windows and for anything wanting tighter time error than 1e-4 in the populations.

**RK4IP composes with the existing adaptive controller.** `adaptive_step!` already implements the Richardson defect estimate and the Söderlind PI3.4 proposal and takes the stepper as a parameter, so all RK4IP needed was a `dt` keyword — no second adaptive path. Measured step sizes on an Eu-like 12³ DDI config, reference error bounded at 2.7e-12 by Richardson on its own method:

| budget | RK4IP dt | Strang dt | speedup |
|---|---|---|---|
| 1e-3 | 6.51e-3 | 4.98e-3 | 1.3× |
| 1e-4 | 3.71e-3 | 1.58e-3 | **2.4×** |
| 1e-5 | 2.11e-3 | 4.98e-4 | **4.2×** |

The payoff grows as the tolerance tightens, as 4th-vs-2nd order predicts. But RK4IP's failure mode is a **wall**: 1.5e-1 relative error at `dt` = 2.5e-2 becomes 4e40 at 5e-2, where the unitary split-step merely degrades. So the controller is a safety property here, not only a speed one. The test starts it past the measured divergence and asserts it pulls back, reaches `t_end`, and lands within 1e-5 of an independent fine-`dt` reference — that last assertion because a loop accepting everything would pass the rest.

**RK4IP is reachable from YAML** as `integrator: rk4ip` on a `dynamics:` step. It was callable from Julia but no config could select it, which is the same defect as a knob that parses and does nothing, from the other side. `_resolve_dynamics_stepper` had no test at all — which is how `integrator:` came to parse and be ignored in the first place — so this adds one that asserts each supported name maps to a **distinct** stepper.

**Two claims in this branch were wrong and are corrected by measurement.**

- `_resolve_dynamics_stepper` said the default is 1st order when `c_dd > 0`. It is not: `split_step!` already dispatches to `_half_potential_step_midpoint!` whenever DDI is active, measured order **2.00 / 2.00 / 1.93** at `c_dd` = 0 / 147.7 / 1477. The 1st-order warning describes the legacy path with `MEANFIELD_MIDPOINT_ENABLED` off. Selecting `midpoint` explicitly is close to a wash: 1.7× lower error at fixed `dt` for 1.49× the cost.
- `rk4ip.jl` claimed "order 2 → 4 at equal cost" from counting mean-field evaluations. Measured ms/step on an Eu-like 12³ DDI config: `split_step!` 3.00, `split_step_midpoint!` 4.46, `rk4ip_step!` **2.61**. RK4IP is *cheaper* than the default. Net of the step it holds: **2.7× at a 1e-4 budget, 4.9× at 1e-5**. CPU only — the GPU ratio is unmeasured and the five scratch buffers are ~2.2 GB at 128³ F64 D=13.

**RK4IP dropped three-body loss and the absorbing boundary**, caught by `test/oracles/test_path_coverage.jl` (every `src` file advancing `ws.state.step` must route dissipation through `apply_rt_dissipation!` or be excused). `_rk4ip_accumulate!` correctly skips `LossTerm` — loss is not Hamiltonian flow — but the real-time epilogue was never added, so a `loss:` config under `integrator: rk4ip` would have run **silently loss-free**. Live rather than hypothetical, since the previous commit made that selectable from YAML. Fixed, with a test that 50 steps at K3 = 5.0 lose norm, a positive control that the same run without loss does not, and a cross-check that RK4IP loses the same amount as `split_step!`.

**Not proposed: making RK4IP the default.** It is not norm-conserving (so `normalize_every` is not honoured), it is real-time only, its failure mode is a wall rather than a slope, and the GPU cost and memory are unmeasured. Opt-in.

**A retracted-and-redone `N_atoms` scan.** The first version overrode `N_atoms` in **one of the three places it lives** (`defaults.interactions`, the `ground_state` block, and by injection into `dynamics`), so its 3.5e4 arm ran a 3.5e4 **ground state** under 5e4 **dynamics** couplings. Caught by a positive control on the cloud size: that arm came out **larger** (rms 3.121 vs 2.948) when a smaller `N` must give a smaller cloud. An out-of-equilibrium state expands.

The irony is the finding — that is precisely the initial-state/dynamics mismatch I had hypothesised in *their* code, reproduced by accident in ours because one knob lives in three places.


anko9801:
## 誤差予算が閉じました — 残り 0.039 nT は**我々の時間刻み**でした

基準: N = 3.5×10⁴、我々の既定で中心 **−2.510**、彼ら **−2.549**。各候補を単独で入れて符号付きで測定（UGE 8313740、両アーム exit 0、45 点ずつ）:

| アーム | 中心 [nT] | 幅 [nT] | Δ中心 |
|---|---|---|---|
| 基準 | −2.510 | 12.89 | — |
| **+ `dt` 4倍細分** | **−2.550** | 13.00 | **−0.040** |
| + 彼らの指数ランプ | −2.587 | 12.90 | −0.077 |
| + 格子（32³/箱16 → 128³/箱36.2） | — | — | −0.007 |
| + DDI カーネル | — | — | −0.002 |
| **Matsui** | **−2.549** | 12.75 | **必要量 −0.039** |

**時間刻みを収束させるだけで −2.550、彼らの −2.549 と 0.001 nT 差。** 彼らも `dt = 1e-3` ですが Crank–Nicolson を自己無撞着まで反復する方式で、こちらは Strang 分割。どちらも 2 次精度で、係数が向こうの方が小さかっただけです。

ランプは **行き過ぎ**（−0.077 対 必要 −0.039）。二つを足すと −0.117 で残差の 3 倍になるので、`dt` の細分とランプの修正は**同じ離散化を部分的に二重に捉えています**。そこまでは分離していません。

## 基底状態の DDI は大差で棄却

`initial.f90` を読んで見つけたもの — 彼らの GS ソルバは `+cdd·INT0_n·MatSZ(im)` を持つのに、**CN 反復内で `INT0` を更新する呼び出しが両方ともコメントアウト**されています。`INT0` は Thomas-Fermi 種から一度計算されたきり凍結。有力候補に見えました。

測ったら違いました。我々の GS の DDI を完全に切ると中心は **−3.205**、ブラケット幅 **0.695 nT**（positive control: GS の形が 0.7786 → 1.3858、裸のトラップ値 1.3967 に一致）。Matsui は**自己無撞着端から 0.039 nT**、反対端からは 0.656 nT。TF 種の双極子場は収束後の場とほぼ同じなので、凍結してもほとんど costless でした。

**中心について未説明の物理は残っていません。**

## 幅は閉じず、未解決として残します

`dt` もランプも幅を**逆方向**に動かします: 12.89 → 13.00 / 12.90 対 彼らの 12.75。中心が 0.001 nT で合ったのに幅は 0.25 nT（2 %）残る。幅は端点ベースラインに対して測る量で、彼らの格子は 1 nT、我々は 0.5 nT なので一部はサンプリングの差だと思われますが、分離していません。

## 最終形

> ¹⁵¹Eu の EdH 共鳴について、独立に書かれた実装が、規約を別々に宣言し結合定数が7桁一致した状態で、Matsui らの公表 Fig. 4B を再現する。原子数を彼らの実行値 `N = 3.5×10⁴` に合わせ、時間刻みを収束させると、ディップ中心は **−2.550 対 −2.549 nT**。幅は 2 % 残り、未解決。

## 作業上の反省を一つ

死んだジョブを 30 分ポーリングしました。ログ最終行が `Step 2/3: DynamicsStep` だったので待ち状態と解釈し、`qacct` の `exit_status` を見ていませんでした。実際は直後に `ArgumentError: enable_ddi=true ... but c_dd not specified` で落ちていた（GS の DDI を切ると dynamics が `ws_prev` から `c_dd` を継承できない）。以降は `exit_status` を必ず併せて見ています。

Issue #250, opened 2026-07-31
Trajectories and cache forks: identity should be a chain, and a factored triple


anko9801:
Two questions came up that the design doc (#237) does not answer, and both change it. Writing them up before implementing anything.

> ある状態から始めてこういう操作をしてこうなった、というのはどうする？
> それと、そういう計算の途中から派生するキャッシュ Fork はどうする？

## 1. A trajectory is not a `Model`

`Model` in #237 is the *resolved physics* — a static description of a Hamiltonian, trap and fields. A real computation is not that. It is:

> start from state `s0`, ramp, quench, hold, measure

That is a **sequence of operations applied to a state**, and it needs its own representation. The proposal:

```
id(s_k) = H( id(s_{k-1}), digest(o_k) )
```

Each step's state identity contains the previous step's. A hash chain.

This is the same structure #237 already proposed for continuation (the B3 B-scan, where link *k* is seeded from link *k−1*). What the question exposes is that **continuation is not a special case — it is just a trajectory**, and the general object is the chain. That simplifies the design: one concept, not two.

## 2. The operation digest has to be factored, or B4 becomes unwritable

First correction the questions force.

`digest(o_k)` cannot be a single hash of everything, because it has to contain both the physics *and* the numerics — `dt`, splitting order, dealias, grid — since those change the numbers too.

But **B4 is exactly "hold the physics fixed and vary the numerics"**: grid {64,96,128} × dt {3} × box {2} × `k_cut` × `ddi_pad_factor` × cutoff on/off. If physics and numerics are hashed together, "same physics, different discretisation" is not a queryable relation and the convergence study cannot be expressed.

So identity should be a triple, not a scalar:

```
id = ( physics_digest ,  numerics_digest ,  input_id )
      which Hamiltonian   how discretised    what it started from
```

Then B4 is the natural query "all artifacts sharing a `physics_digest`, grouped by `numerics_digest`". This is not in #237 and should be.

## 3. Fork from the middle is then free — and that is the strongest argument for per-link identity

Given the chain, forking is structural rather than a feature:

```
s0 → s1 → s2 → s3
           ↘
            s2' = H( id(s2), digest(o') )
```

`s0..s2` keep **the same ids**, so they are shared by construction. Same model as git.

[...]

PR #271, opened 2026-08-01
Cutover step 1: record a complete artifact id beside the key that admits

anko9801:
Step 1 of `docs/design/research_spec_and_provenance_architecture.md` §6 (design in #237, audit trail in #250).

**Admission behaviour is unchanged.** The old `_gs_cache_key` still decides every cache hit; the new `artifact_id` is recorded only. That is what makes this landable on a live tree and revertable by deleting one file. Verified structurally: `experiment.jl` is untouched, and the diff to `run_step_ground_state.jl` is **+8/−0** inside the existing save block with no change to any `isfile` / `cache_path` conditional.

## Lands `src/model/`, and why it needed gates first

`src/model/` existed as an untracked prototype from a pass that was stopped before its gates were written. Landing it required fixing **eight defects it had never been run against**. The first is the whole argument for gating before committing:

> `_trim_pad_factor` was called and defined nowhere, so **every `Model(...)` threw `UndefVarError`.**

"It loads" was true and meant nothing. Also fixed: `active(::DipoleTrapSpec)` compared an SI polarizability (Eu, 5.88e-37) against a *dimensionless* tolerance of 1e-30; polarization normalisation was not idempotent, so decode was not a fixed point; NaN could enter through the five constructors that branch on activity *before* validation; and a `Model`'s hash was mutable through waveform and loss arrays the caller still held.

## New

`Stage` (kind, model, method, from, params, backend), `code_tree_hash()` over `src/` + `ext/`, and `artifact_id(::Stage)` over the existing `content_id`.

The omission rule is a **value** predicate (`_speceq(s, typeof(s)())`), not the physics predicate `active` — the two came apart on `PotentialSpec` and `LossParams`, the two slots with no normalising constructor.

## Gates: 215 assertions, and the canary record is the point

**55 deliberate source corruptions were run.** 50 turned a gate red for the right reason. The **five that came back GREEN were reported as broken gates rather than quietly strengthened**, and all five are now real:

| was vacuous because | repair |
|---|---|
| Julia's `Dict` order depends on table **capacity**, not insertion order — both fixture sides had the same capacity | rebuilt one side through a `sizehint!`-grown table, plus a precondition asserting the orders actually differ |
| the length-prefix fixture used two **single-file** trees, where the separator alone disambiguates | replaced with the 1-file-vs-2-file pair that produces a real collision when the prefix is removed |
| **both `run_registry.jl` writer sites** — the ones every YAML scan uses — were ungated, so this step's deliverable could have been deleted silently | `test_record_provenance.jl` does two real `run_yaml` round trips and gates each site independently |

Two assertions were replaced rather than kept: `Model in roots` was a tautology no source edit could break, and `Model.parameters == Core.svec()` **errored** instead of failing when violated. Both replacements were proven by counterfactual — old and new forms run side by side under the same corruption, old passing, new failing.

## Verification

Five suites, **215/215**, run independently after the formatter pass. **FAST tier green, exit 0**, no failures. Tree verified byte-identical after every canary by checksum over all touched files.

Also corrects one line of the design doc: the `:cuda` dead-code fix was caught in review by reading `backend.jl`, not by a test — luck, not process, which is why step 2 requires its positive control before merge.


anko9801:
## Left for the author: the conflicts are two designs, not two texts

Updated onto current `main` and stopped. Six files conflict, and two of them are genuine design disagreements that I should not settle by picking a side.

**1. `SPIN_TAYLOR_TOL` — should the knob exist?** Both sides independently concluded the shipped 1e-13 default was wrong, and then diverged:

- this branch **deletes** the knob (cutover step 4) and freezes `SPIN_TAYLOR_TOL` as a plain `const`, on the measurement rel|Δψ| = 6.46e-16 at Eu F=6 16³ — 100× under what flipping the realization costs;
- `main` **keeps it registered** at reference == default == 1e-15, arguing "Still registered because it CAN bind at large rotation angles."

Taking `main` silently un-does the cutover step; taking this branch silently drops a documented argument for keeping the knob. Either way a tolerance constant changes without the person who measured it saying so.

**2. Point-cache admission.** This branch adds `admit_payload(psi_file) → (hit, provenance)`; `main` added `_assert_point_provenance(psi_file, env; verbose)`, which `CLAUDE.md` now names as the gate for stale-point reuse. Two implementations of the same guarantee, in the same two call sites.

The other four (`bench/phase_gap_error_budget.jl`, `test/workflow/test_gs_stage_cache.jl`, `test/workflow/validation/test_accuracy_knobs.jl`, and the second `run_registry.jl` hunk) follow whichever way those two are decided.

Nothing was pushed to this branch.

anko9801:
## Measured the premise for the `SPIN_TAYLOR_TOL` half — it does not hold

Cutover step 4 deletes `:spin_taylor_tol` and freezes `SPIN_TAYLOR_TOL` into a `const`, on the grounds that the tolerance cannot bind where this code runs. I went to check that before resolving the conflict, and the premise comes from a number in `src` that is wrong by three orders.

`spin_rotation_taylor.jl` contradicted itself: lines 65 and 87 say production `R ≈ 1e-5`, line 75 says production `R` is 0.01–0.2. Measured — `theta_max = |c1*dt|*fmax*F`, the quantity `spin_mixing.jl:90` already computes, Eu F=6, 16³:

| c₁ | dt | R_max | deg @1e-13 | deg @1e-15 |
|---:|---:|---:|---:|---:|
| −0.05 | 0.005 | 0.00134 | 5 | 5 |
| −0.05 | 0.01 | 0.00268 | 5 | **6** |
| −0.2 | 0.005 | 0.00536 | 5 | **6** |
| −0.2 | 0.01 | 0.0107 | 6 | **7** |
| 1.0 | 0.01 | 0.0536 | 8 | **9** |

The tolerance binds at five of six production cells. Degree 2 needs `R` below ~3e-8.

**Your measurement is not what I am disputing.** `rel|Δψ| = 6.46e-16` is real, and it shows the tolerance does not move the *observable*. The deletion needs the stronger claim that it does not move the *degree*, and that one is false at production — so `code_tree_hash` covering a frozen const would be recording a value that is still doing work.

Corrected in #307, with `test/hamiltonian/test_taylor_tolerance_binds.jl` pinning it so the claim cannot drift back.

## The other half stands as posted

`admit_payload` does not record a producing commit, so it cannot replace `_assert_point_provenance` — the two catch different failures (incomplete/unfit payload vs payload made by different code). Leaning on `admit_payload` alone drops the guarantee `CLAUDE.md` names; leaning on `_assert_point_provenance` alone only forgoes an addition. That asymmetry is why I would keep the provenance check as the gate and add the verdict fields (`converged`, `floor_limited`) rather than swap.

The `floor_limited` distinction is worth having — a BdG or stability gate genuinely should be able to refuse a floor-limited state. That argues for landing the marker machinery *alongside* the commit check, not in place of it.

Still nothing pushed to this branch.

anko9801:
## I tried the resolution I proposed. It does not exist as a merge.

Previous comments recommended keeping `_assert_point_provenance` as the gate and `SPIN_TAYLOR_TOL` as a knob, i.e. resolving both contested points toward `main`. I did that against current `main` (`df4254d0`) to see what came out, and the answer is that it is not a conflict resolution — it is a rejection wearing one.

**`fast` tier: 22 files failed, 10/10 workers.**

- **10 are this branch's own** — `test/model/test_{corpus_resolves,gs_admission_axes,scan_path_admission,resolve_gs_is_shared,no_ambient_module_refs,prose_does_not_move_identity,cache_stats_reported,interrupted_run_recomputes,interrupted_dynamics_recomputes,gs_step_returns_one_state}.jl`. They assert the cutover architecture, so of course they fail when the call sites keep `main`'s.
- **9 exist on `main` too**, and they fail for the same reason from the other direction. `test_cpu_spin_rotation_taylor_parity.jl:153` asserts `SPIN_TAYLOR_TOL == 1.0e-13` — the bare const this branch introduces. `main` never touched that line, so git merged this branch's version with no conflict marker at all, and it then contradicts `main`'s `Ref(1.0e-15)`.

There was also a silent auto-merge defect with no markers: both sides appended the same three kwargs to one `make_workspace(...)` call in `run_step_ground_state.jl:513`, giving `syntax: keyword argument "light_shift" repeated`. The package would not load. Worth knowing generally — a clean `git merge` on this pair is not evidence of a coherent tree.

Nothing was pushed. The working tree is discarded.

## Decision

**This branch should not land in its current form.** Three measured reasons, none of them about effort:

1. **Cutover step 4 is dead.** Deleting `SPIN_TAYLOR_TOL` rests on production being unable to feel it. Measured: production `R` is 1.3e-3…5.4e-2, not the `1e-5` three `src` comments claimed, and the tolerance changes the Horner degree at five of six production cells. Corrected and gated in #307.
2. **`admit_payload` cannot replace `_assert_point_provenance`.** It certifies bytes and now a solver verdict; it records no producing commit. Swapping drops the guarantee `CLAUDE.md` names — that the same YAML under different code cannot silently return old results.
3. **The architecture is woven through ~20 test files and `main` has moved 212 commits**, during which it independently built and shipped the gate this cutover was heading for. The two are no longer alternatives being compared; one is in production.

## What is worth re-landing

The verdict machinery, and only that. `converged` / `stop_reason` / `floor_limited` / `grad_norm` carried beside the cache entry, so a BdG or stability gate can refuse a floor-limited state — a real distinction, resting on this repo's own measurement that `tol = 1e-8` sits under the ~5e-7 gradient floor, so `converged = false, floor_limited = true` is the best achievable answer rather than a failure.

That is a small PR **on top of** current `main`, adding to `_assert_point_provenance` rather than replacing it. It keeps the part of `src/model/complete.jl` that pays for itself and drops the cutover framing that no longer has a target.

Closing this is the author's call, not mine — but I would not spend more on the merge.

anko9801:
Closing. Both halves of the cutover are answered, and neither answer is "merge this".

- **Step 4, deleting `SPIN_TAYLOR_TOL`** — the premise is false. Production `R` is 1.3e-3…5.4e-2, not the `1e-5` three `src` comments claimed, and the tolerance changes the Horner degree at five of six production cells. Measured and gated in #307.
- **The cache-admission cutover** — `admit_payload` records no producing commit, so it cannot stand in for `_assert_point_provenance`, which `main` has since shipped as the documented gate. Attempting the merge with the call sites kept at `main`'s left 22 files red across 10/10 workers, plus a marker-free auto-merge defect that stopped the package loading.

The branch stays; nothing was pushed to it. `src/model/complete.jl` is not thrown away by this — the piece worth keeping is the solver verdict (`converged` / `stop_reason` / `floor_limited` / `grad_norm`) carried beside the cache entry so a BdG or stability gate can refuse a floor-limited state. That belongs on current `main` **alongside** `_assert_point_provenance`, as a much smaller PR, and without the cutover framing that no longer has a target.

Reopen if the verdict work is better done by reviving this than by re-cutting it.

PR #276, opened 2026-08-01
The defect catalog on TSUBAME: nothing escapes the full tier, and one 17 s file guarded a fatal class alone

anko9801:
`test/mutation/` (#198) has 58 mutants, 35 of them `:fatal`, and had only been
run over `test/workflow/`. Which defect classes ship green was unmeasured. This
adds the job script, runs the whole catalog, and fixes the three ways the
measurement was misreporting itself.

## The answer

**Against the whole `full` tier with the cost cap lifted, 0 of 58 escape.** Six
escaped the default `grounded_cheap` / 15 s probe and all six were caught on
re-probe — `workflow/test_lhy_block_wiring.jl` (256 s) is one of the catchers and
the cap is what removed it.

What survived is a tier question. Every escapee has only **one or two** catchers,
so where those files live decides whether a PR would see the defect. Required is
`fast` + `oracles` + `integration`:

| mutant | severity | caught by | required? |
|---|---|---|---|
| `yaml_ddi_padding_default_off` | gross | `test_config.jl`, `test_yaml_physics_reaches_workspace.jl` | ci ✓ |
| `yaml_lbfgs_drops_lhy_table` | fatal | `test_lhy_block_wiring.jl` + 1 | fast ✓ |
| `dealias_k_cut_half` | subtle | `test_dealias_2_3.jl` only | ci ✓ |
| `b_block_cartesian_direction_guard` | gross | `test_b_block_normalize.jl` only | fast ✓ |
| **`yaml_calibration_not_applied`** | **fatal** | **`workflow/test_pipeline.jl` only** | **✗ FULL_EXTRA** |
| **`save_every_off_by_one`** | major | **`workflow/test_pipeline.jl` only** | **✗ FULL_EXTRA** |

`workflow/test_pipeline.jl` costs **17 s** and was the single point of failure for
a fatal class: `p_mv`/`coil_mode` no longer resolving to Gauss before parsing,
which puts every downstream number off by the coil calibration factor while the
run looks entirely normal. Moved to `CI_EXTRA`, which is what `INTEGRATION_TESTS`
derives from.

## Three fixes to the instrument, each found by its own output

**A missing verdict was read as a red test.** `run_probe` returned `Bool` and
mapped "no marker" onto `FAIL`. Safe in the escape direction, and in the other it
credits a dead worker with a catch — booking a real gap as covered and never
re-probing it. Now `:green` / `:red` / `:unknown`, with UNRESOLVED for "nothing
caught it but the probe was incomplete". It fired at once:
`oracles/test_continuity_equation.jl` came back `:unknown` in 4 of 4 shards, and
passes in isolation on the same commit and node type through the same code path
(`PASS 9.48`, job 8315002).

**"A real gap" was a claim about the suite from a measurement about the probe.**
That is the label I misread from this output before re-probing, and the nightly
(#275) runs exactly that default probe — so its escape list will keep naming
covered classes. Renamed to "Caught by nothing in the probe", and the section now
carries its own boundary and the command that settles it.

**The seconds were placeholders.** `min_cover` is sold as "the evidence-based
`fast` tier" and minimised `_cost`, which is `_DEFAULT_COST = 3.0` for everything
outside its 88 hand-entered files, measured on a 4-vCPU runner. The smoke's
"3 files, 9s" was 3 × placeholder and the greedy tie-break ran on it. Meanwhile
`run_chunk.jl` writes `<PASS|FAIL> <secs> <file>` every pass, so a 58-mutant
sweep measures each probe file 59 times on the machine that schedules it — and
`run_probe` read the verdict out of that string and dropped the seconds beside
it. Now kept, median, and marked as measured or estimated.

## The job script, and the two ways it refuses

**A skipped test is green, so it catches nothing.** On a node without CUDA every
GPU-gated file skips and its mutants come back escaped — a report full of gaps
that do not exist. So the script asserts `CUDA.functional()` in its own process
rather than trusting the queue name.

**The harness edits `src/` in place**, so the shared project root is refused by
path; the sweep gets its own checkout. Also uses #272's preflight instead of
restating the BLAS pin, keeps 58 precompiles on the node-local depot, and names
the likely cause on failure (137 = OOM) while re-checking `git diff -- src`.

`qsub -v` cannot carry a command line — it splits the value on commas **and**
spaces, so a `--mutants a,b,c --probe x` knob makes qsub try to open `b` as the
job script. Knobs are separate space-free variables, assembled in the script.

## Caveats

- Every coverage count is a **lower bound**. Stage 2's baseline excluded
  `test_quality.jl` and `workflow/test_vtk_export.jl`; stage 1 intermittently
  excluded the two GPU=CPU parity files. All four are green in this PR's CI —
  they are GPU/host contention from 8 CUDA contexts on one `gpu_h` share, not
  defects.
- The positive control (`spin_mixing_c1_sign`, known caught) was carried in all
  six jobs and caught in all six, so a shard reporting gaps because its julia
  died would have shown up there.
- Jobs: 8314740 (smoke), 8315020-8315023 (stage 1), 8315814-8315815 (stage 2),
  8315002 (the continuity isolation probe).


anko9801:
## Smoke measured (job 8314740, 3 mutants, `grounded_cheap`)

The instrument works, and the cost is now known rather than guessed.

| | |
|---|---|
| probe pool | **264 files** (grounded methods, tier-listed, ≤ 15 s) |
| wall clock | 18:51 → 19:44, i.e. baseline + 3 mutants = 4 passes in **53 min** |
| **per pass** | **~13 min**, dominated by the per-mutant `Pkg.precompile()` |
| result | all 3 fatal sign mutants caught; **0 escaped** |
| GPU guard | `CUDA.functional() = true` — the probe was a real probe |

Catches, for the record:

- `zeeman_b_to_p_sign` → 4 files
- `zeeman_quadratic_sign` → 3 files
- `spin_mixing_c1_sign` → 7 files

## Two things the smoke surfaced that are not about this PR

**`oracles/test_continuity_equation.jl` is RED before any mutation** and is
excluded from the catch matrix — so 1 of 265 files silently contributes no
coverage to every sweep. It is `CI_EXTRA`, i.e. inside the `oracles` tier, and
that tier **passes on this very branch** (5m54s, above). So it is red under the
harness and green in the suite, which is a difference in the harness, not a
defect in the file. Job 8315002 is probing three candidates in isolation
(direct include ± `SPINORBEC_FFT_ESTIMATE`, and via `run_chunk.jl`).

**34 files were dropped over `--max-cost=15` (2719 s total)**, including
`oracles/test_spin_chain_fusion_parity.jl` and `workflow/test_lhy_block_wiring.jl`.
The harness names them rather than cutting them silently, and stage 2 lifts the
cap for whatever escapes.

## Full sweep launched

58 mutants at 13 min/pass is ~11 h serially, so it is split across 4 jobs
(8314996-8314999) with **4 separate checkouts** — the harness mutates `src/`, so
sharing a tree would have them compiling each other's defects. Largest shard is
17 mutants + baseline, ETA ~3.9 h.

Each shard carries `spin_mixing_c1_sign` as a **positive control**: it is known
to be caught by 7 files, so a shard that reports it as escaped has an
environment failure, not a coverage finding. Without that, a shard whose julia
died early would report 14 fatal gaps that do not exist — which is the same
failure this PR's GPU guard exists to prevent, one level up.

anko9801:
## Result: nothing escapes the full tier, and one file that guards a fatal class ran in no required check

Two stages, 58 mutants. **Stage 1's six "escapes" were artifacts of the cheap
probe — against the whole `full` tier with the cost cap lifted, 0 of 58 escape.**
Retracting the "33 % of the config layer ships green" reading I posted from stage
1 alone; it was measured against a 264-file grounded/≤15 s pool, not against what
the suite contains.

| stage | probe | escapes |
|---|---|---|
| 1 (8314740, 8315020-23) | `grounded_cheap`, `--max-cost 15` → 264 files | 6 / 58 |
| 2 (8315814-15) | `dir:` = whole `full` tier, `--max-cost 400` | **0 / 6** |

`workflow/test_lhy_block_wiring.jl` (255.9 s) is one of the catchers and stage 1
had dropped it over the cost cap, which is most of the explanation.

## What survives, and it is the actionable part

Every escapee is caught by **one or two** files, so the question is which tier
those files are in. Required is `fast` + `oracles` + `integration`:

| mutant | severity | caught by | required? |
|---|---|---|---|
| `yaml_ddi_padding_default_off` | gross | `test_config.jl`, `test_yaml_physics_reaches_workspace.jl` | ci ✓ |
| `yaml_lbfgs_drops_lhy_table` | fatal | `test_lhy_block_wiring.jl`, `test_yaml_physics_reaches_workspace.jl` | fast ✓ |
| `dealias_k_cut_half` | subtle | `test_dealias_2_3.jl` (only) | ci ✓ |
| `b_block_cartesian_direction_guard` | gross | `test_b_block_normalize.jl` (only) | fast ✓ |
| **`yaml_calibration_not_applied`** | **fatal** | **`workflow/test_pipeline.jl` (only)** | **✗ FULL_EXTRA** |
| **`save_every_off_by_one`** | major | **`workflow/test_pipeline.jl` (only)** | **✗ FULL_EXTRA** |

`workflow/test_pipeline.jl` costs **17 s** and was the single point of failure for
a fatal class: `p_mv`/`coil_mode` stopping their resolution to Gauss before
parsing, which puts every downstream number off by the coil calibration factor
while the run looks entirely normal. Moved to `CI_EXTRA` (`INTEGRATION_TESTS`
derives from it), verified `INTEGRATION` 38 → 39 files with no cross-list
duplicate.

## Instrument findings, both fixed in this PR

**A missing verdict was being read as a red test.** `run_probe` returned `Bool`
and mapped "no marker" onto `FAIL`, which is safe in the escape direction and
credits a dead worker with a catch in the other — booking a real gap as covered.
Now `:green` / `:red` / `:unknown`, and a mutant nothing caught while files went
unreported reports **UNRESOLVED**. It fired immediately:
`oracles/test_continuity_equation.jl` came back `:unknown` in **4 of 4 shards**,
and it passes in isolation on the same commit and node type through the same code
path (`PASS 9.48`, job 8315002).

**The seconds were placeholders.** `min_cover` is sold as "the evidence-based
`fast` tier" and minimised `_cost`, which is `_DEFAULT_COST = 3.0` for everything
outside its 88 hand-entered files. The smoke's "3 files, 9s" was 3 × placeholder,
and the greedy tie-break ran on it. Meanwhile `run_chunk.jl` writes
`<PASS|FAIL> <secs> <file>` every pass, so a 58-mutant sweep measures each probe
file 59 times on the machine that schedules it — and `run_probe` read the verdict
out of that string and dropped the seconds. Now kept, median, and marked.

## Caveats on the numbers above

- Stage 2 baseline excluded `test_quality.jl` and `workflow/test_vtk_export.jl`
  (red on TSUBAME, both green in this PR's CI `fast`), and stage 1 excluded
  `oracles/test_gpu_cpu_per_term_parity.jl` / `test_level0_gpu_cpu_consistency.jl`
  intermittently. Those are GPU/host contention on a `gpu_h` share — 8 CUDA
  contexts on one GPU — not defects. So every coverage count is a **lower bound**.
- The positive control (`spin_mixing_c1_sign`, known caught) was carried in all 6
  jobs and caught in all 6. A shard reporting gaps because its julia died would
  have shown up there.
- Stage 2 had **zero** `:unknown` at 6 workers.

Issue #281, opened 2026-08-01
Re-derivation: all four figures of the four-figure spec moved


anko9801:
All four figures of `docs/manuscript/four_figure_spec_2026_05_26.md` have now been
re-derived on current `main`. **Every one moved.** The qualitative conclusions
mostly survive; not one of the numbers does.

Figures 1 and 3 are written up in `docs/campaign/` (#257, #274). Figure 4 is below
and not yet committed. Figure 2 was re-derived by an earlier session.

---

## Why these four and not the other 477

Of the 481 stored `summary.json` files, **zero are reproducible from a commit** —
18 carry `_repo_commit` and all 18 are `-dirty`. But they are not equally
load-bearing:

| | run dirs |
|---|---:|
| holding a stored summary | 474 |
| cited by a live document **by exact name** | **10** |
| matched only by a family glob (`runs/klaus_*`) | 72 |
| cited by nothing at all | **392** |

Reading the context of those 10 (rather than counting them): **8 back a figure or a
validation claim row**, two are config-location pointers. All ten are dated
2026-05-26/29 and unstamped — the vintage the audit classifies as predating every
correction.

The four figures are where those 8 concentrate.

---

## Figure 1 — peak density 16.5 % lower, grid axis irrelevant (#274)

`l4_k3_ladder` n64 rung + `matsui_baseline` n64. Comparing `peak_max` to
`peak_max`, both time-maxima from each run's own `dynamics/peak_density`.

| config | stored | re-derived | change |
|---|---:|---:|---:|
| `L4_K3_n64_HamOnly` | 0.00951101 | 0.00793751 | **−16.5 %** |
| `L4_K3_n64_K3` | 0.00949813 | 0.00792665 | **−16.5 %** |
| `L4_K3_n64_K3gdr` | 0.00949780 | 0.00792608 | **−16.5 %** |
| `L4_K3_n64_gdr` | 0.00951079 | 0.00793694 | **−16.5 %** |
| `L4_K3_n64_K3LHY` | 0.00953759 | 0.00692943 | **−27.3 %** |
| `matsui_40ms_dynamics_n64` | 0.00716601 | 0.00612421 | −14.5 % |
| `matsui_5ms_morphology_n64` | 0.00611232 | 0.00452326 | −26.0 % |

Four of five L4 cells move by the *same* −16.5 %, agreeing to 0.14 %: K3 and γ_dr
change nothing. `K3LHY` is the outlier and is the only cell with LHY active, so
[...]

PR #288, opened 2026-08-01
The disposal count was measuring the regex: a citation gate that stopped at the first path segment

anko9801:
Closes two of the three items left in #281. **Nothing is deleted here.**

## Classification

474 run directories hold a stored `summary.json`.

| class | dirs | |
|---|---:|---|
| **exact name** | 10 | a live document names the directory |
| **glob family** | 72 | a live document names `runs/<stem>_*` |
| **uncited** | 392 | nothing in `docs/**` outside `archive/` |

### The 10 exact-named were the four figures — and all four are now re-derived

8 back a figure or a validation claim row. **Every figure resting on them has been
re-run on current code, and every one moved** (Fig 1 −16.5 %; Fig 2 premise gone;
Fig 3 seven of ten classifications; Fig 4's comparison turned out not to be a
symmetry operation).

So these ten are no longer *evidence* for anything — they are the record of what
was believed before the corrections. A real use, and a much smaller one.

### The 72 glob-matched back no quantitative claim

| stem | dirs | cited by |
|---|---:|---|
| `klaus_*` | **65** | one research-log entry — the "data source" column of a table recording **eight retired Python figure scripts** |
| `sprint5_M1_*` | 4 | a data-layout example in `architecture/sweep_view.md` |
| `eu151_*` | 2 | a cookbook example |
| `eu_k3_*` | 1 | matsui status |

Sixty-five of seventy-two are provenance for a *deletion*.

### 7 of the 392 are tracked in git

`eu_collapse_search`, `eu_ham_only_conservation`, `eu_ham_only_ramp_quench_24`,
`matsui_edh_baseline_{529e3a77,9ca97308}`, `saito_li_torus`,
`yan_li_saito_f1_grid_refinement`. Someone committed these deliberately —
**not-cited is not grounds for disposal on its own.**

## Footprint — why this is worth doing

| location | uncited dirs | disk |
|---|---:|---:|
| `BEC-simulation` (main checkout) | 148 | **~230 GB** |
| `wt:silly-foraging-flame` | 213 | **~199 GB** |
| three other worktrees | 38 | — |

**Over 400 GB**, against a 2 TB group quota already 1.2 TB used.

## The disposable set: 385

392 uncited − 7 tracked. Each satisfies all four of: no live citation by name or
glob; untracked; no producing commit (like all 481, not reproducible); predates the
2026-06/07 corrections so any number in it is superseded by construction.

## Procedure: `mv`, not `rm`

The classification is evidence, not proof, and a document could cite one of these
tomorrow — at which point `test_doc_run_citations_resolve.jl` says so and the
directory comes back. Deleting forecloses that; moving does not. **Re-measure
before running it**: `docs/` changes daily and a new citation moves a directory
out of the set.

Docs only.

### DIFF
--- a/docs/campaign/rederivation_fig1_l4_ladder.md
+++ b/docs/campaign/rederivation_fig1_l4_ladder.md
@@ -113,6 +113,31 @@ a single `N_trajectory` block (`Fz_per_N`, `DeltaFz`, `N_final_ratio`, …);
   summaries contain `N_trajectory` and nothing else, so there is no like-for-like
   density number to put beside the panel's table.
 
+## The inset survives where the main panel does not
+
+The inset's two lossy arms were re-run on TSUBAME. Its quantity is the surviving
+fraction N(T)/N(0), which the stored summaries record as
+`N_trajectory.N_final_ratio` and the v3 extractor as `1 − norm_rel_drift` — the
+same number by construction, unlike ΔF_z below.
+
+| config | stored | re-derived | change |
+|---|---:|---:|---:|
+| `matsui_40ms_lossy_medium` | 0.516038 | 0.516901 | **+0.17 %** |
+| `matsui_40ms_lossy_strong` | 0.244302 | 0.243647 | **−0.27 %** |
+
+**Two to three tenths of a percent, against −16.5 % on the same figure's main
+panel.** The inset does not need redrawing.
+
+That is a discrimination rather than a reassurance: both panels went through the
+same pipeline on the same day and only one moved. Peak density is a local maximum
+of the field and rides on the corrected LHY and dealias paths; the surviving
+fraction is a volume integral of a K3 decay and does not. A campaign that had
+re-derived only the inset would have concluded the figure was fine.
+
+The re-derived runs also classify — `stable_arrest` (medium),
+`sacrificial_arrest` (strong) — where the stored summaries classify nothing. New
+information, not a comparison.
+
 ## Not covered
 
 - Fig 1 (d)'s density slice is an analysis of `matsui_5ms_morphology_n64`, which
--- a/docs/campaign/rederivation_fig1_l4_ladder.md
+++ b/docs/campaign/rederivation_fig1_l4_ladder.md
@@ -77,8 +77,47 @@ loss does.
 either dataset — but every number under it moved by 16.5 %, and the figure should
 be redrawn rather than re-captioned.
 
+## Fig 1's inset: the two lossy arms
+
+Re-run on TSUBAME (`gpu_h`, 64³, `import CUDA` first). The inset plots
+N(t)/N(0), so the comparable quantity is the surviving fraction at 40 ms.
+
+| config | stored `N_final_ratio` | re-derived `1 − norm_rel_drift` | change |
+|---|---:|---:|---:|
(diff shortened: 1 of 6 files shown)


anko9801:
## The count was not a property of the data

anko pushed twice on the number — first *"at least keep the phase diagrams"*, then *"that's still a lot."* Chasing the second produced the finding that matters more than any of the numbers:

**385 → 280 → 259 → 207.** The disposable set moved every time the citation matcher was fixed, and **every fix moved directories out of the set, never in.** Three defects:

1. **Nested citations read at their first segment only.** `runs/verification_suite/L4_eu_matsui_hamiltonian_only_*` resolved to `verification_suite`, which exists. The 13 directories it names — called *"the canonical Eu Hamiltonian-only prediction for this codebase"* in `self_contained_validation_report.md:107` — scored uncited and reached the manifest.
2. **Brace expansion not expanded.** `day_inventory_2026_05_26.md:61` cites `L4_K3_n{64,96,128}_*_<hash>`; only the `n64` arm matched. The `n96`/`n128` arms are **91 GiB, 55 % of the bytes then marked disposable**.
3. **Only `runs/`-prefixed tokens counted.** That citation is a bare directory name in a table cell.

Defects 1 and 3 were also in `test/oracles/test_doc_run_citations_resolve.jl`. Fixing it exposed ~30 previously-invisible unresolved citations, and turned up a real one: **nothing named `L4_eu_matsui_hamiltonian_only_*` has ever existed under `verification_suite/`**, in this worktree or the main checkout. Three documents pointed at a path that was never there. Fixed here.

## The gate was checkout-dependent

`runs/klaus_quench/summary.json` exists in the main checkout and not in a worktree, because configs are tracked and outputs are not. Marking those `(gone)` from a worktree would have written a falsehood into the documents. Two changes:

- a citation ending in a file is adjudicated at its **parent**;
- `(untracked)` joins the marker vocabulary for citations naming an output directory.

Resolution is now per path and family-aware — glob metacharacters are captured whole instead of truncating the name, and a family is checked against the longest literal prefix in its parent. Canaried over tracked paths only, so it means the same thing in a worktree, in main and in CI.

## What the manifest now recommends

A list produced this way is fine for *"stop citing these as evidence"* and **is not fit to drive `rm`.**

207 uncited, 82 still on disk, **60.6 GiB against a 261 GiB `runs/` tree**. Reclaim by bytes, not by count — the largest 5 are 38 % of it, 29 of the 82 are under 100 MiB:

| GiB | directory |
|---|---|
| 9.44 | `LHY_polar_contact_200ms_d82e14b5` |
| 4.73 | `LHY_polar_contact_100ms_283e1a72` |
| 3.36 | `L4dealiasv6_kcut11_eu_matsui_hamiltonian_only_96_f7d8dbf9` |
| 3.35 | `L4dealiasv4_eu_matsui_hamiltonian_only_96_6d8d0a7a` |
| 2.42 | `LHY_polar_contact_50ms_3e2af5d3` |

Short enough to read one directory at a time, which is the only review this evidence supports. Still `mv`, not `rm`.

`L4_K3_n{64,96,128}` is a grid-convergence ladder whose name contains none of the words the "keep the scans" rule looks for. It survived because defect 2 was found first, not because the rule caught it.

PR #295, opened 2026-08-02
Two claims were struck through over data that is sitting on disk

anko9801:
Follow-on to #288. The citation inventory's **regenerate-or-retract list** was built with the matcher whose three defects that PR documents. Re-adjudicated with the fixed resolver against the main checkout, **two of its four entries resolve**:

| run | verdict | what is actually there |
|---|---|---|
| `lhy_mode_ablation` | absent | — |
| `twa_sinatra` | absent | — |
| `sprint5_M1_multistart_groundstate` | **on disk** | 30 `cell_B*_Om*.jld2` + `groundstate_audit.jld2`, 83 MB |
| `ddi_convention_factorial` | **on disk** | `results.jld2`, 10 885 bytes, 2026-05-26 |

Both are untracked — invisible from a clone, from CI and from a worktree, present where the numbers were produced.

Acting on the old list, two documents retracted claims they did not need to retract:

- `m1_groundstate_audit_2026-06-08.md` — *"どの commit にも存在せず、下記の数値は再チェックできない … 入力の 30 セルが無い"*, of a sweep that is **entirely there**.
- `self_contained_validation_report.md` Layer F — struck through the output line and told the reader the numeric rows *"cannot be re-checked"*.

Reading **untracked as gone** is the same conflation that put the canonical Eu Hamiltonian-only runs on a disposal list, and it is why the gate now carries an `(untracked)` marker.

Layer F's substantive point is unchanged and restated rather than dropped: the authority is the test, which runs on every push — not any stored number. The correction is about what the reader is told is *available*, not about what is *asserted*.

Regenerate-or-retract is **2, not 4**.

### DIFF
--- a/docs/campaign/doc_run_citation_inventory.md
+++ b/docs/campaign/doc_run_citation_inventory.md
@@ -78,7 +78,29 @@ lost; the citation points at the pre-move path. Repointing it is a doc fix.
 `12174e883326ecac`). These describe a tree the reader does not have. They mislead,
 but they assert no result.
 
-### Live claims with a stated result — 4
+### Live claims with a stated result — 4, of which 2 were false alarms
+
+> **Correction (2026-08-02): two of these four are on disk.** This list was built
+> with the matcher whose three defects are documented in
+> [`stored_run_disposal.md`](stored_run_disposal.md). Re-adjudicated with the
+> fixed resolver against the main checkout:
+>
+> | run | verdict | what is actually there |
+> |---|---|---|
+> | `lhy_mode_ablation` | absent | — |
+> | `twa_sinatra` | absent | — |
+> | `sprint5_M1_multistart_groundstate` | **on disk** | 30 `cell_B*_Om*.jld2` + `groundstate_audit.jld2`, 83 MB |
+> | `ddi_convention_factorial` | **on disk** | `results.jld2`, 10 885 bytes, 2026-05-26 |
+>
+> Both are **untracked**, so they are invisible from a clone, from CI and from a
+> worktree — but the numbers they back can be re-checked in the main checkout.
+> The two documents that had struck their claims through on the strength of this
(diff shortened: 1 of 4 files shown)


anko9801:
## The general question, measured: 249 of 251, not 32

The two false retractions turned out to be the visible edge of a bigger miscount. Asking the general form — *for each run directory holding data, does any committed artifact name it?* — gives:

| | dirs |
|---|---:|
| recipe reachable in git (tracked YAML, or a script/test naming it) | **249** |
| nothing in git names it | **2** — `dipole_field_demo`, `yan_li_saito_f1_torus_gs_disc`, 0.1 GiB, cited by no document |

**That measurement moved 219 → 166 → 62 → 2**, and every step was a defect in the question, not a change in the tree:

1. **"Recipe" was assumed to mean a YAML in the run directory.** Most of this work is driven by committed *scripts* — `scripts/m1_b1_multistart_newton.jl`, `scripts/validation/run_validation_matrix.jl`.
2. **Directory names matched whole**, so every CAS name was lost to its `_<hash>` suffix. Stem matching recovered **104**.
3. **Only file *contents* were grepped, not tracked *filenames*.** The factorial arms are committed at `runs/eu_robust_factorial/K0_gdr0_LHY1.yaml` — a name no content grep surfaces. Recovered **60**.

Same failure as the citation matcher in #288, a **fourth time in one session**: one assumed form for the target, everything outside it reported absent, the count moving one way only.

This also strengthens the correction this PR already makes: both `sprint5_M1_multistart_groundstate` and `ddi_convention_factorial` have a **producer committed**, so they are regenerable and not merely re-readable. **Neither retraction was warranted on any reading.**

**What this does not soften.** A script that names a directory does not pin the commit, the parameters or the Julia version — exactly the gap #290 closes for `summary.json` going forward. The narrow claim is that the stored tree is not an orphanage: it is 2 directories and 0.1 GiB.

PR #312, opened 2026-08-03
bench: L-BFGS lands on the ITP dt→0 limit — and energy converges at twice the state order

anko9801:
Closes the highest-value open item from #308 §C, and corrects two things about how orders were being read.

## The prediction, and it holds

ITP is GFDN, so its converged state carries an **O(dt)** error (2.4e-2 in density at the shipped `dt`, measured in #306). **L-BFGS has no `dt`** — the `dt = 0.001` at `lbfgs/driver.jl:201` only fills a `SimParams` so `make_workspace` can be built; the solver line-searches rather than propagating. So E_LBFGS should sit at the `dt → 0` **end** of the ITP sequence, not beside any particular `dt`.

Eu F=6, 64³, `c1_ratio = +0.05`, `polar_contact`:

| arm | E | wall |
|---|---|---|
| ITP dt = 2e-3 | 771.31593366 | 163 s |
| ITP dt = 1e-3 | 771.07534001 | 203 s |
| ITP dt = 5e-4 | 771.01122439 | 406 s |
| **L-BFGS** | **770.99038657** | **101 s** |

Differences 0.24059 / 0.06412 → ratio 3.75 → **p = 1.91**. Richardson at that order gives **E(dt→0) = 770.98793**, and L-BFGS sits **3.2e-6 relative** from it — with its own floor at `|∇E| = 7.22e-5` (it stops at the energy-comparison floor, and the warning names `residual_polish=true` as the way below, unused here) plausibly accounting for the residual.

Both falsification conditions written into the bench *before* the run were checked and neither fired: L-BFGS is not above the dt/4 point, and not far below the extrapolation. It is also **4× faster** than the ITP arm it beats.

**One grid, one c₁, one seed**, and at the production `c1_ratio = −0.024` the shipped `dt` diverges outright, so the comparison cannot be posed there without first choosing a stable step.

## Correction 1: the extrapolation used the wrong order, in the direction that fails a correct prediction

The Richardson step was hard-coded to the first-order form `2E(h/2) − E(h)` because GFDN is a first-order splitting. That is true of the **state** and false of the **energy**:

| quantity | order |
|---|---|
| density (state) | p ≈ 0.93 |
| total energy | **p ≈ 1.91** |

Assuming p = 1 put `E(dt→0)` at 770.947 instead of 770.988, making L-BFGS look **5.6e-5** from the limit instead of 3.2e-6 — a factor of 17, and in the direction that reports a correct prediction as failing. The order is now **computed from the sequence**.

Same category as comparing a halving ratio to 4 when the target is `2ᵖ + 1` (#306): an order assumed rather than measured turns a textbook result into an anomaly, or the reverse.

## Correction 2: energy converges at TWICE the state order, so it is a poor probe of state accuracy

The two orders above are not in conflict. **E is stationary at the minimum**, so a state error `δ` costs an energy error `O(δ²)` — and `2 × 0.93 = 1.86 ≈ 1.91`.

The consequence is practical: **an energy that looks converged sits on a state whose error is an order larger.** At the shipped `dt` the ITP energy error is 4.0e-4 while the density error is 2.4e-2 — **sixty times larger**. Accuracy claims have to be quoted on a state-level observable, not on energy.

## Also fixed: `save_every` silently gates the ITP convergence check

`_run_itp_loop!` puts `tol`, `tol_drho` and the assignment of `final_dE` inside `if step % sp.save_every == 0` (`itp_loop.jl:160`). The `save_every = 10^9` used here to suppress output **turned the convergence check off**, and every arm returned `conv=NO` with `dE=NaN` for a state whose energy matched a previously converged run **to seven digits**.

The failure is silent in the plausible direction: `converged = false` reads as "ran out of steps", which invites raising the cap. `dE = NaN` from ITP means **never measured**, not diverged.

## What this implies

For accuracy, **"use L-BFGS" beats "make ITP faster"** — it is closer to the `dt → 0` limit than any ITP run here and 4× faster than the one it beats. That is an uncomfortable answer to a session about ITP speed, and it is one measurement at one point; the production sign remains unchecked.


anko9801:
## Correction: the speed ratio is withdrawn; the accuracy result reproduced exactly

A second run of the same bench, after the `save_every` fix, gives **identical iteration counts and energies to eight digits** — and a completely different wall time for one arm.

| | run 1 (`r18n8`) | run 2 (`r4n3`) |
|---|---|---|
| L-BFGS, 1000 iterations | **47.4 s** | **167.9 s** |
| ITP dt/4 | 406 s | 350 s |
| L-BFGS total | 101 s | 356 s |
| E_LBFGS | 770.99038657 | 770.99038657 |

So **"L-BFGS is 4× faster than the ITP arm it beats" is withdrawn.** Within run 1 (both arms on `r18n8`) the ratio is 4×; within run 2 (both on `r4n3`) it is ~1×. Putting both arms on one node is *not* enough to make the comparison sound — L-BFGS line-searches through host BLAS level-1 and is machine-dependent (156.6 → 50.2 ms/it measured previously on the same problem), while ITP is GPU-FFT bound, so a node's CPU/threading state moves one arm and not the other.

**The accuracy result is unaffected and reproduced bit-for-bit**: E_LBFGS = 770.99038657 both times, `|∇E| = 7.22e-5` both times, **3.2e-6 relative** from the Richardson limit at the measured order p = 1.908.

Also confirmed by the rerun: with `save_every = 200` the ITP arms report `conv=yes` with `dE ≈ 3.5e-12`, and their energies match the earlier run to 8 digits — so the previous `conv=NO / dE=NaN` was exactly the suppressed-check artifact described in the PR body, not a real non-convergence.

**What still stands:** for *accuracy*, L-BFGS sits at the ITP `dt → 0` limit while every ITP run at a practical `dt` does not. **What no longer stands:** any claim about it being cheaper. That needs a repeated, same-node, both-arms measurement, and it is not in this PR.

PR #316, opened 2026-08-04
Matsui Fig. 2C: reproduced at rms 0.0075, and the m=-6 residual is in the transfer, not the loss

anko9801:
Three results and one main-branch bug fix.

## 1. The reproduction holds across the whole time series (type A)

`runs/matsui_fit/fig2c_n35k.yaml` at their conditions — N = 3.5e4, B = 2.6 nT,
40 ms, loss-free. Against their published simulation, sampled every 0.5 ms:

| component | rms |
|---|---|
| m = −6 | 0.0063 |
| m = −5 | 0.0093 |
| m = −4 | 0.0072 |
| m = −3 | 0.0078 |
| m = −2 | 0.0063 |
| **all five** | **0.0075** |

The criterion was written into the config before launch (< 0.02 establishes it,
> 0.05 says the conditions are mismatched). An earlier arm at N = 5e4 sat at
0.109, so the entire gap was the atom number.

Third independent dataset to agree, after the 45-field dip (per-field rms 1.1 %)
and the 5 ms column density (0.3–4 % on five observables). `N = 3.5e4` now has
six supporting observables.

## 2. Correction: the comparison was normalising away the answer

`dataset_fig2_exp.csv` carries **absolute atom numbers** — rows sum to 4.478e4 at
t = 0 and 2.811e4 at 40 ms, the paper's 38 % loss. Normalising each row to its
own sum divides that decay out.

| m = −6 vs their simulation, 1–40 ms | rms |
|---|---|
| as a fraction of the surviving atoms | 0.2633 |
| as a fraction of the initial number | **0.0942** |

Same data, same simulation. The 0.263 this arc set out to explain was 2.8×
inflated by the denominator, and the component reported as worst is the one that
agrees best.

Where the lost atoms went needed no run at all: of the 0.4297 deficit at 40 ms,
**0.0 % is in m = −6** — the experiment holds 0.0236 *more* m = −6 than the
loss-free model. That is the paper's own statement made quantitative. The doc
flags the arm of this that is a tautology (the total deficit equals the measured
loss identically) so it does not get quoted as evidence.

CLAUDE.md gains the rule this bought: a normalisation is a measurement choice,
print the row sums of any external fixture once, and re-derive a
single-component discrepancy under a second normalisation before explaining it.

## 3. The calibrated loss explains m = −4 and m = −2, and not m = −6 (type C)

`runs/matsui_fit/fig2c_loss.yaml`, `K3 = 2.6e-28 cm^6/s` on the twelve m ≠ −6
components, calibrated to the paper's 26 %/40 ms. No free parameters.

| component | loss-free | with loss | |
|---|---|---|---|
| m = −6 | 0.0926 | 0.0986 | **−6 %** |
| m = −5 | 0.0473 | 0.0443 | +6 % |
| m = −4 | 0.1497 | 0.1148 | **+23 %** |
| m = −3 | 0.0497 | 0.0476 | +4 % |
| m = −2 | 0.1667 | 0.1153 | **+31 %** |

Exactly the paper's split: m = −4 and m = −2 are the two it names as loss-free
casualties, and they are the two the loss moves. m = −6 gets worse — loss raises
its *fraction* but lowers its *count*, because atoms leave m = −6 by transfer and
only become vulnerable afterwards. A larger K3 widens that gap.

So the m = −6 residual is in the **transfer**, with the same sign as the Fig. 4B
residual: the simulation drives the cascade out of m = −6 harder than the
experiment does.

The pre-recorded criterion is filed as **invalidated, not passed** — it asked
m = −6 to fall below rms 0.10 and the loss-free run was already at 0.0926, so it
could not fail. It was written when that number was believed to be 0.263.

## 4. `energy_gradient!` dispatched on the wrong array

It chose CPU vs GPU from `psi` while computing on `ws.state.psi` and writing
`grad`. A ground state read back from jld2 is a host `Array` whatever backend
wrote it, so the GS stage-cache audit — host ψ, GPU workspace — took the CPU
branch, where the registry's device buffers met a host `grad` and every
accumulate scalar-indexed. The audit could never run on GPU.

It needs a cache **hit** to surface: the run that fills the cache computes its
ground state and never reaches the load path, so only the *second* GPU run of a
given ground state dies. That is how it killed a Fig. 2C arm here.

Both faces (`energy_gradient!`, `gradient_only!`) now key on `ws.state.psi` and
require `grad` to be co-resident rather than scalar-indexing to an answer;
`verify_verdict` allocates its scratch from the workspace. Gated by
`test/gpu/test_gpu_energy_gradient_host_psi.jl` (full tier), which asserts
GPU == CPU *value* rather than "does not throw" and covers the production
`verify_verdict` shape including that it still rejects a wrong energy. Canaried:
reverting the branch key reproduces the scalar-indexing error.

## Not tested here

Whether the transfer rate itself is over-driven — `c1_ratio`, `q`, or the
dipolar drive. Every arm in this campaign so far fixed those at the values their
simulation ships, so none of them could have seen it.


anko9801:
Coverage note on the GPU gate: `test/gpu/test_gpu_energy_gradient_host_psi.jl` is in the `full` tier, and neither the per-PR jobs nor nightly run on a GPU runner — `CUDA.functional()` is false there, so **the file is a no-op in all CI**. It was verified on the local WSL2 card (12 passes) and canaried by reverting the branch key, which reproduces the production `Scalar indexing is disallowed` at `_grad_kinetic!`. The same is true of every other `gpu/` file, so this is not a new hole, but the green checks on this PR say nothing about that fix.

The four `xfer_*` sensitivity arms (tasks 50–53, job 8347696) are still running on TSUBAME; their verdict will be added to `docs/validation/matsui_fig2c_absolute_numbers.md` before merge.

anko9801:
**Transfer-rate arms are in, and the answer is that nothing there is resolvable.** Sections 6–7 of `docs/validation/matsui_fig2c_absolute_numbers.md`.

Six arms, m = −6 rms against the experiment in absolute number (baseline 0.0926):

| arm | m = −6 |
|---|---|
| r = 1/72 | 0.1408 |
| r = 1/9 | 0.1292 |
| q = 0, B = 2.6 nT | 0.0726 |
| q = 10 Hz | 0.3105 |
| q = 0, B = 1.92 nT | 0.1142 |
| q = 0, B = 3.28 nT | 0.1080 |

**One clean positive: `c1_ratio` is excluded.** Worse in *both* directions across a factor of 8 in `r`, so their 1/36 is a local optimum for this observable. Consistent with the ring-count bound and with `c1_ratio` not moving the Fig. 4B dip. A two-sided point needs no error bar.

**`q` looked like the answer and is not measurable here.** The last two arms bracket q = 0 across the ±0.68 nT that q/h = 1 Hz shifts the resonance by. m = −6's rms moves **0.0416 over 1.36 nT of field**; the `q` effect being interpreted is **0.0200, i.e. 48 % of that**; and the paper quotes **up to 10 nT of offset error** on the field axis — 7.4× the range scanned. The pre-recorded criterion resolves to neither branch it anticipated, and the reading is the third: a single-field time series cannot constrain any of these parameters.

Gate 3 arrived one step late. The systematic was written down for the *dip* comparison, where it is handled right — a dip **width** is invariant under a field offset, which is why the *centre* was never quoted as agreement. A single-field cut inherits the offset with nothing to average it out. CLAUDE.md gate 3 now says the systematic binds the axes held **fixed**, not only the one being scanned.

What stands unchanged: the offset-invariant comparison. Our width and theirs agree to 0.10 %, both miss the experiment's by 14–15σ, residuals correlated at 0.9705 — their model against their experiment, with no parameter of ours in it.

Incidental confirmation of the GPU fix: both bracket arms loaded a cached ground state and passed the stage-cache audit. On `main` they would both have died at `_grad_kinetic!`.

PR #322, opened 2026-08-04
docs(campaign): what 2026-08-04 established beyond the physics

anko9801:
The physics verdict already has a home — `docs/validation/matsui_campaign_report.md`.
This adds the part that belongs to no subsystem, in a durable form:
`docs/campaign/lessons_2026_08_04.md`, linked from `CAMPAIGN.md` and from the
campaign report.

Kept because **five of the six corrections were invisible from the inside**: each
mechanism was healthy and each answer was confidently wrong.

**One latent code defect.** `energy_gradient!` branched CPU-vs-GPU on `psi` while
computing on `ws.state.psi`, so the GS stage-cache audit could never run on GPU. It
needs a cache **hit**, and the run that fills the cache never loads from it — so only
the *second* GPU run of a given ground state dies. A defect on a cache-hit-only path
is invisible to any test that builds its own fixture. (Fixed in #316; recorded here
for the shape.)

**Two defects in the instruments, not the work.** An agent has no clock, so a
promised completion report needs a process whose *exit* supplies the turn (#319). And
`MEMORY.md` was 43 % over its 24.4 KB load limit, so **41 % of the index had not been
loaded in any session** — undetectable from inside one, because the warning names the
file and not the entries that fell off the end.

**The six corrections, with classes.** Two generalise beyond this campaign:

- *The regex became the measurement* — fired three times in one day (a retracted-arm
  count that matched live files citing them, a `git diff` across two divergent trees,
  and a `grep -E "^[+-][^+-]"` that excluded its own target lines). When a correction
  always moves in one direction, the rule has become the measurement.
- *Normalisation vs held-fixed systematic* — the two ways healthy machinery returns a
  self-consistent wrong number: one projects out the degree of freedom under test,
  the other reads a difference smaller than the uncertainty on a coordinate being
  held still.

Filed **FROZEN**: the findings are about events and are permanent; the paths in §5
are as of today. `test_docs_live_set.jl` caught the missing marker on the first run —
55/55 after.

For the record on what the gates bought: 2026-08-02 was 24 arms and ~3 h GPU with
every premise wrong; 2026-08-04 was 6 arms and 36 min, each with its rejection
criterion written into the config before launch, and the negative result
(`c1_ratio` is excluded) is the useful one. Same question, same code.


anko9801:
Campaign closing record: #323 — carries the verdict and tracks the two open items.

anko9801:
## main is red, and this PR now also fixes it

Landing this exposed a live breakage on `main`, from the same class of hole that was closed on 2026-07-31.

**#321 added `docs/design/lbfgs_speed_limits.md` with neither a `LIVE_DOCS` entry nor a `FROZEN` header** — and `test_docs_live_set.jl`, the gate that forbids exactly that, never ran on it. Its four checks were green; two of them were **skips**: `Test (tier=fast)` "passed" in **10 s** and `Integration` in **8 s**, because a docs-only diff sets `code=false`. Only the oracles job, which honours the `docs` output, actually ran.

The 2026-07-31 fix added that `docs` output because `test/oracles/test_doc_run_citations_resolve.jl` reads `docs/**.md`. `test_docs_live_set.jl` was added later to the **fast** tier, whose steps are gated on `code` alone — so the gate went unreachable from the only change that can break it, **one tier over from where the hole was patched**.

Two commits added here:

- the `FROZEN` header `lbfgs_speed_limits.md` needs — main green again, 55/55
- the fast tier's steps now honour `code || docs`, so **any** docs-input test in that tier is reachable, not just this one. A docs-only PR pays ~6-8 min for it; the previous arrangement bought that time back by not running the gate.

Note how this was found: my own PR was green and main was green, and the **combination** was red — the shape of #227/#228. It surfaced only because I merged main before merging rather than after, which is worth keeping as the habit.

Verified: YAML parses, `test_docs_live_set` 55/55, `test_tier_membership` green (it reads the tiers back out of `ci.yml`, so the edit had to survive that too).

anko9801:
**Correction: the fast-tier timing I cited is not evidence.**

I reported `Test (tier=fast)` running 7m38s here as showing the fix works. It does not — this PR touches `.github/workflows/ci.yml`, so `code=true` and that tier runs either way.

I then opened #325 as a positive control with a `.md`-only *commit*. Also invalid, for a reason worth recording: **the filter reads the PR's `base..head` diff, not the last commit** (`ci.yml` lines 82-84). #325 branched off this one, so it carried the `ci.yml` change and `code` was true again. The tell was in its own run — `Integration gates` ran 13m12s, and integration's steps were never given the `docs` condition, so `code` must have been true. Closed.

What is actually established:

- `docs` is computed correctly — #321's oracles job ran 6m31s on a docs-only diff, which requires `docs=true`
- the change here is a plain `|| needs.changes.outputs.docs == 'true'` on the four fast-tier steps

So the fix is sound by inspection, and the one untested link is that OR. A valid control needs the fix on `main` first and then a PR whose **entire** diff is `.md` — I'll run that immediately after this lands and report the fast job's duration either way.

Recommending merge on that basis: `main` is red now, and the `FROZEN` header in this PR is what makes it green.

anko9801:
Control passed: #326 — a PR whose entire diff is one `.md` file ran `Test (tier=fast)` for **9m58s** while `Integration gates` skipped in **11s**. The skip is the negative control in the same run: it proves `code=false`, so the fast tier ran on `docs` alone. Pre-fix, #321's fast job was 10s.

The gate is now reachable from the only change that can break it.

PR #325, opened 2026-08-04
[control, do not merge] docs-only diff must run the fast tier

anko9801:
Positive control for the CI fix in #322, **not for merging**.

The 7m38s `Test (tier=fast)` on #322 proves nothing about that fix: #322 touches `.github/workflows/ci.yml`, so `code=true` and the tier would have run either way. This branch changes **only a `.md` file**, so `code=false` and `docs=true`.

- **fast tier runs (minutes)** → the fix works
- **fast tier reports ~10 s** → it does not

Will be closed and the branch deleted once it reports.


anko9801:
**Invalid control — closing.** It measured nothing.

The filter reads the PR's **base..head** diff (`ci.yml` lines 82-84), not the last commit. This branch carries the `.github/workflows/ci.yml` change itself, so `code=true` and the fast tier would have run with or without the fix.

The tell was in the same run: `Integration gates` also ran (13m12s), and integration's steps were never given the `docs` condition — so `code` must have been true.

A valid control needs the fix already on `main` and a PR whose **entire** diff is `.md`. Will run that after #322 lands.

PR #326, opened 2026-08-04
[control, do not merge] valid docs-only diff must run the fast tier

anko9801:
Positive control for the CI fix that landed in #322. **Not for merging.**

#325 was invalid: it branched off the PR carrying the `ci.yml` change, and the filter reads the PR's `base..head` diff rather than the last commit, so `code=true` and the fast tier ran regardless.

This branches from `main` with the fix already in. Its **entire** diff is one `.md` file.

- **fast tier runs for minutes** → the `|| needs.changes.outputs.docs` condition works
- **fast tier reports ~10 s** → it does not, and the docs gate is still unreachable from the only change that can break it

Closed as soon as it reports.

### DIFF
--- a/docs/campaign/lessons_2026_08_04.md
+++ b/docs/campaign/lessons_2026_08_04.md
@@ -149,3 +149,6 @@ Neither is a computation.
 
 Anything measured at a single field, against a residual smaller than the field
 systematic, reproduces §5 of the campaign report.
+
+<!-- Positive control 2026-08-04: a PR whose ENTIRE diff is .md must still run the
+     fast tier, because test_docs_live_set.jl lives there and reads docs/**.md. -->


anko9801:
**Control passed — closing.**

Docs-only diff (`base..head` = one `.md` file):

| job | duration | reading |
|---|---|---|
| `Test (tier=fast)` | **9m58s** | **ran** — the `\|\| docs` condition works |
| `Integration gates` | **11s** | skipped, correctly — its steps are `code`-only |

The negative control is inside the same run, which is what makes it conclusive: `integration` skipping proves `code=false`, so `fast` ran on `docs` alone. Against #321 before the fix:

| | fast | integration |
|---|---|---|
| #321 (pre-fix, docs-only) | **10s** skipped | 8s skipped |
| #326 (post-fix, docs-only) | **9m58s ran** | 11s skipped |

`test_docs_live_set.jl` is now reachable from the only change that can break it.

PR #330, opened 2026-08-05
fix(evap): robust ramp optimizer + T₀/K₃/α anchoring (partially superseded by main)

anko9801:
> **Correction, added after opening this PR.** I first described the whole of
> this branch as new. It is not — roughly the rate-form half of it is already
> on `main`, landed independently while this worktree sat abandoned. The
> overlap is spelled out below, and **this PR should be reduced to the part
> that is genuinely novel before it is reviewed as a unit.**

Recovered from an abandoned worktree (first commit 2026-07-24). Two commits.

## Already on `main` — do NOT re-review these as new

| this branch | already on main |
|---|---|
| 3-D LRW rate with the ≤0 clamp | `evaporation_model.jl:158` — `max(eη * (η*P3 - 4*P4) / P3^2, 0.0)` |
| `c(η) = 3 P(4,η)/P(3,η)` mean energy | `evaporation_model.jl:145` |
| `evap_scale` removed | superseded on main; the audit ledger records the earlier `evap_scale ≈ 0.02` fit as dead |
| condensate `⟨n²⟩` 4/7 → 8/21 | `a5c7db4d` |
| figure driver consolidation | `0e146a9e` (this is why two figure drivers conflict add/add) |

That is also why the merge conflicts: `src/solvers/evaporation/{evaporation_model,condensate}.jl` collide because both sides changed the same physics, and main's side is current.

## What is genuinely not on main

- **`src/solvers/evaporation/evaporation_robust.jl`** (195 lines) — worst-case optimization over the calibration uncertainty set. `main` has no such file. The aggressive schedule sits near two cliffs — the loaded-depth floor, where `η_start < 4` means no ramp evaporates at all, and the 3-body cliff, where ~2× the assumed `K₃` misses BEC while the gentle lab ramp still reaches it — so a schedule to hand the lab should be robust rather than optimal.
- **The `T₀` epoch re-anchor, 50 µK → 18 µK.** The 2022 PRL temperature was being paired with the 2023 notebook's ramp — two different generations of the apparatus. That epoch-mixing drove `η_start ≈ 2`, which is unphysical (real ODTs load at η ~ 5–10) and inflated the fitted `K₃` to compensate. At 18 µK, `η_start ≈ 5.7`. No `18e-6` anchor exists on main.
- **`K₃` as ab-initio rather than fitted** — universal van-der-Waals `K₃ = 3C ℏa⁴/m` (Braaten–Hammer) instead of `1.61×10⁻⁴⁰` fitted to the measured BEC number. Across the `C ∈ [0,67]` band the predicted `N_BEC` spans ~5×10⁴…10⁶ and **brackets** the measured 5.02×10⁴, which sits at the high-loss edge. Eu is only marginally universal (`a/r_vdW ≈ 1.3`), so this is order-of-magnitude and is stated as such rather than as a match.
- **`α` off a trap-frequency calibration and onto the Eu ⁸S₇/₂ atomic constant** `5.88×10⁻³⁷ J/(W/m²)` (≈189 a.u.).

## Verification type

**C (model fidelity)** for the endpoint, in its honest form: a prediction band that brackets the measurement, not a tuned identity. **A** for the convention fixes (`test/solvers/test_evaporation.jl`, `test_evaporation_tools.jl`) — though the `⟨n²⟩` half of that is main's already.

## Suggested disposition

Do not merge as-is. Either:

1. **Reduce to the novel set** — `evaporation_robust.jl` + the `T₀`/`K₃`/`α` anchoring, rebased on current `main`, dropping the rate-form changes that main already has; or
2. **Close** and re-land the two novel pieces as their own PRs.

I have not picked between these, because which side of `evaporation_model.jl` is current is a physics judgement and main has been moving in this file for two weeks (`2ce0218c`, `a7cd1391`, `bc0b03ec`, `80355942`, `05b96ec2` are all SPGPE work on top).

Also note ~65 of the 89 files are rendered figures, their CSVs and their generators, all regenerable from the drivers here.


anko9801:
Closing. The code should not land; the diagnosis was worth having and is now verified and written down elsewhere. Everything below was measured on current `main`, not read off this branch.

## The centre of this PR — `evaporation_model.jl` — should not be taken

Four findings, in order of how decisive they are.

**1. The removal of the `1/√2` rests on a statement that is false.** This branch changes `γ_el = n σ v̄/√2` to `n σ v̄`, justified in the code comment as *"no extra 1/√2 (v̄ = √(8k_BT/πm) is already the mean relative speed)"*. It is not: `√(8kT/πm)` is the mean speed of one particle, and the mean **relative** speed for equal masses is `√2 ×` that. `main`'s `n σ v̄/√2` equals `½ n σ v̄_rel`, the per-atom rate counting each pair collision once, and the repo states that form in two places — `docs/guides/evaporation_model.md:30` and `docs/research_notes/evaporation_bec_prep_model_2026-06-15.md:232`, the latter while arguing the rate is fully determined. The change is a factor √2 on the evaporation rate with no re-validation.

**2. The `V_ev ≤ 0` clamp never fires.** It is described here (and in this PR's body, which I wrote) as "the first-principles replacement for a hard η floor". `V_ev = η P(3,η) − 4 P(4,η)` is positive and monotonically increasing across the whole tested range — `1.87×10⁻⁴` even at η = 0.5. The claimed mechanism is dead code.

**3. The cooling law disagrees with `main`'s by up to 3.2× exactly where forced evaporation runs, and the asymptotic check cannot arbitrate.** Both forms converge to O'Hara's `(η−2)/3` at large η (at η = 20: `main` 1.0055×, this branch 0.9966×), so the limit test passes on both sides. The disagreement is confined to η ≲ 8:

| η | `main` `L` | this branch `L` | ratio | `(η−2)/3` |
|---|---:|---:|---:|---:|
| 4.0 | 0.821 | 2.638 | **3.22×** | 0.667 |
| 5.7 | 1.345 | 2.246 | 1.67× | 1.233 |
| 8.0 | 2.081 | 2.375 | 1.14× | 2.000 |
| 12.0 | 3.388 | 3.353 | 0.99× | 3.333 |

This branch's `L` is also **non-monotonic**, with an interior minimum at η = 6.24 — in the middle of the ODT loading window — where `main`'s form and the reference are both monotonic. Its `max(denom, 0.1)` floor is honest and binds only below η ≈ 1.5. The **eject factor is bit-for-bit identical** on both sides, so that part of the refactor contributes nothing.

**4. Taking this file reverts two features `main` has since gained**: `noneq_scale` (the non-equilibrium finite-evaporation-rate penalty, which is what bounds "faster is always better" into a real interior optimum in ramp duration) and `as_mult` (Feshbach-tuned `a_eff` with `K3_eff = K3·as_mult⁴`).

## Corrections to this PR's own body

Two claims I put in the body were wrong and are withdrawn:

- **"`evap_scale` removed → superseded on main."** `main` still has the knob, defaulting to `1.0`. What was superseded is the fitted *value* `≈ 0.02`, not the parameter.
- **"`α` moves to the atomic constant."** `_EUV3_ALPHA = 5.88e-37` is already on `main`; that was not this branch's contribution.

## The diagnosis WAS right, and it is now landed

`η_start` at the shipped defaults is **2.07**, and `docs/guides/evaporation_model.md` claimed `≈ 4.5` in four places. That is a real defect, it is this branch's finding, and it is fixed in #333.

## …but the replacement anchors do not close the endpoint

Measured on `main`:

| | `T₀` | `K₃` | `η_start` | `N_BEC` / measured |
|---|---|---|---|---|
| `main` shipped | 50 µK | 1.61e-40 | **2.07** | 1.31× |
| this branch | 18 µK | 1e-41 | 5.75 | **10.62×** |
| `T₀` only | 18 µK | 1.61e-40 | 5.75 | 1.47× |
| `K₃` only | 50 µK | 1e-41 | **2.07** | 3.42× |

And the "ab-initio band brackets the measurement" claim does not hold. Sweeping `K₃ ∈ [10⁻⁴²·³, 10⁻³⁹·³]`, the measured `5.02×10⁴` is reached only at `K₃ ≈ 2.45×10⁻⁴⁰` (T₀ = 50 µK) or `2.74×10⁻⁴⁰` (T₀ = 18 µK) — both **2.5× above the band top** `1×10⁻⁴⁰`. The whole band over-predicts. That is model deficiency, not a parameter choice, and it is also recorded in #333.

## Why the robust optimizer is not being salvaged either

`evaporation_robust.jl` (195 lines, no conflict) was the one clean addition, but: it is **not self-contained** — it calls `_descend_monotone_family`, which lives in this branch's `evaporation_optimize.jl` refactor (55 pass / 2 error when ported alone); `main` already hedges α and K₃ via `param_uncertainty_ensemble` + `optimize_ramp_monotone(; ensemble=)`, so what it adds is the *operational*-error axes (power, imbalance, T₀, N₀, timing); and optimizing the robustness of a model whose endpoint needs a `K₃` 2.5× outside the physical band is out of order. The branch stays on the remote, so reviving it later costs nothing.

Also, if it is revived: `scale_K3` forwards only five of `EvapParams`' fields, so a caller who set `noneq_scale` or moved `evap_scale` would have scenarios silently evaluated under different physics than the nominal ramp — which is not a control.

## Not carried over

The ~65 files under `docs/guides/figures/` are rendered figures, CSVs and their generators, regenerable from the drivers here.

PR #331, opened 2026-08-05
Six statements in CLAUDE.md were false, and the memory audit could not see its own rot

anko9801:
One theme, thirty-two commits: **an instrument or a document reporting a state
the tree does not have.** It started as six false lines in `CLAUDE.md`. Deriving
those counts instead of fixing them turned up the same shape in live code — a
condition testing a vocabulary that had moved, a reader opening a path nothing
writes, a gate measuring the form production never takes — and each of those got
a fix plus a gate that fails if it comes back.

69 files, +4935/−211. Nothing here changes physics: every numerical difference
below is a case where the code was **not** doing what its own configuration
said.

## 1. Documents that were wrong about the tree

`CLAUDE.md` is loaded into every session, so each false line is paid by every
future agent. Statements were checked against a `file:line` or a sha before being
changed, and found false in three passes — six, then eleven counts, then a final
six. Representative ones:

| claim | reality |
|---|---|
| inner-V forward chain lists 7 operators | `_outer_operators_fwd!` issues **9** — `spatial_lhy_spin`, `spatial_zeeman` missing |
| "the 3 sibling converters delegate to it" | four files, seven call sites |
| `hamiltonian/{shared,optics}/` | relocated by `5355e455`; neither exists |
| external Ueda is `BLOCKED_EXTERNAL` | `ueda_status.md:112` disclaims the token |
| procedure lives in `adding_new_hamiltonian_term.md` | `2fb43967` reduced it to a pointer back here — a round trip |
| memory section cites `scripts/audit_memory_refs.py` | deleted in the session that added the citation |

**The operator list had a root cause in the code.** `split_step.jl:541` carries
a hand-maintained copy of the function body twenty lines below it; the copy was
short by the same two operators and `CLAUDE.md` had faithfully copied it. The
2026-08-04 sweep corrected 5 → 7 and landed two short for exactly this reason.
Both are fixed and the comment now says it is a copy and names the function as
the authority.

Three more documents, each wrong in a way a reader would act on:

- `docs/reference/dynamics.md` described what `tol` bounds incorrectly — a LIVE
  reference, so it is now derived.
- Three YAML analyzer names in the guides **never existed**, and one measurement
  was quoted with its sign pointing the wrong way.
- `yaml_schema_reference.md` cited `runs/option_gamma_*/config.yaml` as what the
  ground-state arms protect. No such directory exists; the pin is the test.

**Counts are now derived, not typed.** `scripts/generate_state.jl` (new, 994
lines) emits `docs/STATE.md` from the tree — tier file counts, the artifact
identity key set, the cache-admission ratio (4 sites admit, 1 verifies), the
`m_lbfgs` default split — and `test_state_doc_is_current.jl` fails when the
committed copy drifts. A stale enumeration is exactly how a wrong-sign B→p
converter survived two months, so the converter count is a `grep` now, and
`test_bfield_sign_declared_once.jl` gates that **exactly one** line computes the
sign (`units.jl:73`, unchanged). Five new gates sit around documents:
`CLAUDE.md`'s own citations must resolve or name their ref, examples may not use
removed keys, analyzer names must name real implementations, the state doc must
match its deriver, and the B→p sign must be declared once.

A caveat worth stating: a derived document is only as good as its deriver.
`generate_state.jl` shipped once with a regex reading `CI_EXTRA` as 90 against a
true 113 — a non-greedy match stopping at a `]` inside a comment. It evaluates
the file now, and its floor gate was widened, because a lower bound catches a
collapse and not a plausible-looking corruption.

## 2. `scripts/audit_memory.py` — the gate reported its own blindness as a result

- `INDEXES` was a hardcoded pair; **eight** index files exist → 24 false orphans.
- `reachable` followed only `[[wikilinks]]`; sub-indexes use `](name.md)`, so
  every sub-index contributed zero children.
- `load` never read `archive/` → 74 false broken index links.

True counts after the fix: **0 unreachable, 0 broken index links.** Added under
one judgement-free rule — act only where the resolution is unique, otherwise
report and leave it: an `ON ANOTHER REF ONLY` class (a path on another branch is
not deleted), `--fix-relocations` (9 of 45 dead paths were moves),
`--fix-links`, a DRIFT/FORWARD-REF split so the gate stops failing on
spec-legal content, wider path coverage (`figs/` alone is 43 tracked files), and
a calibration guard that refuses to report when `MEMORY.md` reaches nothing —
because a broken extractor and a clean tree print the same words.

## 3. Code that did not do what its configuration said

Each has a gate named after the defect.

**Integrator / physics paths**

- `epsilon:` sized `dt` for an integrator that never runs on the rotating path —
  **10⁴× the requested tolerance**. (`test_dt_matches_the_integrator_that_runs.jl`)
- The DDI path realises a composition as an un-merged triple jump, not the
  merged ABA product the order gate measured. `omelyan_pefrl` and
  `blanes_moan_srkn6b` are order 4 as ABA and **order 2** as the jump — four to
  six stages per step for the order plain Strang already gives, on every
  production Eu run. Now refused rather than silently substituted; no config
  under `runs/` selects either. (`test_composer_order_conditions.jl`)
- The absorbing mask ignored the step it was applied to.
  (`test_absorbing_boundary_honours_the_step.jl`)
- Padded DDI runs left `ws.ddi_bufs` at zero and two readers took that as a
  field. (`test_ddi_bufs_are_empty_when_padded.jl`)
- `dynamics.adaptive_dt` was validated and then discarded; inert keys are
  refused now. (`test_inert_dynamics_keys_are_refused.jl`)

**Autopilot / cluster**

- VRAM sizing under-estimated **16 of 23 atoms, by up to 5.67×** — it consulted
  a second, stale atom table. Both that and the budget gate now read
  `ATOM_REGISTRY`. (`test_profile_vram_uses_the_registry.jl`,
  `test_no_second_atom_F_table.jl`)
- A GPU OOM never reached `RESOURCE_PERMANENT`, so retry re-ran the recipe
  instead of escalating the resource class — two defects, stacked.
  (`test_oom_reaches_resource_permanent.jl`)
- The GPU preflight **could only ever say yes**. (`test_preflight_can_fail.jl`)
- Failure evidence was read from `run_dir/stderr.log`; `LocalBackend` writes
  `.autopilot/stderr.log`, so the stderr arm never matched and every local
  failure fell through to "no evidence".
  (`test_failure_evidence_reaches_the_reader.jl`)
- Every Slack alert was a no-op and no caller could tell.
  (`test_slack_alerts_report_delivery.jl`)
- The divergence kill covered one of three dynamics paths; all three now report
  liveness through one writer.
  (`test_every_dynamics_path_reports_liveness.jl`)

**Dashboard / analysis**

- Frame indices were served under a key the front end plots as time.
  (`test_dashboard_does_not_invent_a_time_axis.jl`)
- The vector-field plan cache was keyed on resolution alone, so a different box
  reused the plan. (`test_plan_cache_is_keyed_on_the_box.jl`)
- Transverse spin on the rotating path was a literal `0.0`, written out as a
  measurement. (`test_transverse_spin_is_measured.jl`)
- The `full_bdg` cost advisory tested `initial_state: ferromagnetic`, retired in
  favour of `m_plus_F`/`m_minus_F` — a false negative on exactly the configs it
  exists for, which paid the full BdG cost with no hint that the closed forms
  agree to ~1e-4. Gated as a property over the enum, not a name list.
  (`test_full_bdg_advisory_fires.jl`)

`test/helpers/calibrated_scan.jl` is the shared instrument behind several of
these: a scan helper that cannot return a number it has not first proved it
could get wrong.

## Verification

The five files that were red on the previous head were re-run individually
first: state doc 42/42 · doc-run citations 7/7 · `full_bdg` advisory 9/9 ·
autopilot 219/219 · split_step 48/48. Four of those five reds were this branch
extending something and not updating its reader — including a
`composition::NamedTuple{(:a, :b)}` kwarg that stopped accepting its own
module's constants once they gained `order`/`jump_order`.

JuliaFormatter pinned at 2.5.0, clean on `src/` and `test/`. `scripts/` is
outside what CI formats and was already drifted at the merge-base — not touched,
to avoid reflowing unrelated lines.


anko9801:
## 7件目 — `even_c_extra` は存在しない関数を canonical として名指ししている

この PR の diff にこの2行が **context 行として通過**しているので、表に足す候補として。

- `CLAUDE.md:245` — "Even-rank only via `even_c_extra(F; c2, c4, c6, …)`"
- `CLAUDE.md:259` — "**`even_c_extra(F; c2, c4, c6, …)` is canonical**"

**この関数はツリーに存在しない。** `.git` を除く全文検索でのヒットは3箇所だけで、
どれも定義ではない:

| 場所 | 何 |
|---|---|
| `CLAUDE.md:245`, `:259` | canonical API としての名指し（本件） |
| `.claude/scripts/hook_post_write.sh:43` | この名前を含むファイルを警告する grep パターン |
| `src/model/specs.jl:178` | docstring が "`even_c_extra`'s documented misindexing footgun" と参照 |

実体は

```julia
interaction_params_from_constraint(; c_total, c1_ratio, F, c_extra::Dict{Int,Float64})
```

（`src/hamiltonian/coefficients.jl:222-255`、`c_extra[4] = c₄` のようにランクをキーにする Dict）。
YAML 側は `_parse_higher_rank_c_n`（`src/workflow/experiments/schema/parsing_blocks.jl:397`）が組む。

**注意すべき点**: CLAUDE.md が警告している footgun そのもの（手書きの `[c2, c4, c6]` が
F ≥ 3 で silently misindex する）は正しく、しかも **Dict でランクをキーにする現行 API では
構造的に解決されている**。誤っているのは解決策として挙げられた関数名だけ。だから
「警告を消す」のではなく「実在する API 名に直す」のが筋で、`specs.jl:178` の docstring も
同じ名前を参照しているので併せて。

失敗クラスはこの PR が `scripts/audit_memory_refs.py`（それを引用したセッションで削除された）
について指摘しているものと同型 — **存在しないものを canonical として名指しする**。

依存マップの issue #342 の副産物として見つけたもので、そちらにも記録しています。

